PR #2, opened 2015-09-17
Seg padding

tjmassin:
Tested the output of the code with "-s 0" as command line argument and results were unchanged. Also tested leaving out the -s argument at all and results were unchanged. Just to make sure, ran with a large padding that removed one trigger but not the one just before it and it worked well.

Added an if statement that says, "If my segment is shorter than my padding window, go to the next segment because data query would fail"


duncanmmacleod:
thanks @tjmassin

PR #19, opened 2016-07-07
Update gwdetchar-slow-correlation to show Pearson's coeff 

isapatane:
Changed lines 109-111 and 256-257 to include the Pearson's coefficient for the range versus the darm blrms.

### DIFF
--- a/bin/gwdetchar-slow-correlation
+++ b/bin/gwdetchar-slow-correlation
@@ -253,7 +253,7 @@ numpylink = html.markup.oneliner.a(
 page.p("In the results below, %s is calculated as the square of %s using %s."
        % (rsq, wikilink, numpylink))
 if args.trend_type == 'minute':
-    page.p("%s is reported for both <code>%s</code> and <code>%s</code>. The ",
+    page.p("%s is reported for both <code>%s</code> and <code>%s</code>. The "
     "%s between these two channels is %.2f." % (rsq, primary, rangechannel, rsq, corr0))
     
 page.div(class_='panel-group', id_='results')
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@isapatane, it looks like there's an error on line 256 of your edited `gwdetchar-slow-correlation`:

```python
    page.p("%s is reported for both <code>%s</code> and <code>%s</code> The %    s between these two channels is %.2f."
            % (rsq, primary, rangechannel, rsq, corr0))
```

should probably be

```python
    page.p("%s is reported for both <code>%s</code> and <code>%s</code> The "
                "%s between these two channels is %.2f." % (rsq, primary, rangechannel, rsq, corr0))
```

or similar, whatever works. Please review [PEP8](https://www.python.org/dev/peps/pep-0008/
), specifically the guidelines about line length, sorry, I'm pedantic about this sort of thing.

duncanmmacleod:
@isapatane, please post an example command-line use of this executable, so I can test the new code end-to-end.

isapatane:
command line: `(gwpysoft) [isa.patane@ldas-pcdev1 L1-slow-correlation-1133906417-7199]$ gwdetchar-slow-correlation -i L1 -b 45 90 -f chans.txt 1133906417 1133913616`
Used this code to display https://ldas-jobs.ligo-la.caltech.edu/~isa.patane/detchar/O1/L1-slow-correlation-1133906417-7199/

isapatane:
Okay it should be good now. Right after I posted the command line I went back and realized it was not exactly the same as the version I had been testing but i just double checked and it should be correct now.

PR #24, opened 2016-07-27
Scaling data with RMS to correct for outliers within the data

katerinaleman:
The scaling of data using the max values of the data failed to work when there were any big deviations in any of the channels. Here is an example of the output display that the original version of the code creates when there is an outlier in the data.
image-1

Isa created a short Python notebook where she uses arbitrary functions as two different sets of data to show how  the scaling of data using RMS works much better when having a big deviation in the data. (https://files.slack.com/files-pri/T04CYEMQN-F1V0EU49X/detrendwithrms.pdf)

After incorporating the RMS scaling to the code (by making local changes to the code) I ran the code for the same time that the code showed to be faulty up above in the picture and these are the new results. (https://ldas-jobs.ligo-la.caltech.edu/~katerin.aleman/detchar/03/L1-slow-correlation-1133906417-14443/)

### DIFF
--- a/bin/gwdetchar-slow-correlation
+++ b/bin/gwdetchar-slow-correlation
@@ -132,9 +132,12 @@ corr3 = spearmanr(rangets.value, darmblrms.value)[0]**2
 # create scaled versions of data to compare to each other
 print("-- Creating scaled data")
 rangescaled = rangets.detrend()
-rangemax = rangescaled.value.max()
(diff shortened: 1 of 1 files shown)


katerinaleman:
@duncanmmacleod : This is the command line I used to test for the changes in my pull request. `(gwpysoft) [katerin.aleman@ldas-pcdev1 L1-slow-correlation-1133906417-14443]$ gwdetchar-slow-correlation -i L1 -b 45 90 -f chans.txt 1133906417 1133920860`

PR #42, opened 2016-11-18
Threshold update

uberpye:
THINGS DONE (numbered by the updated code's lines):
at about 179: added a threshold variable
at about 200 to 204: if the all the correlations are below a threshold it does not plot
at about 329 to 331: marks that channel as below the threshold
at about 337 to 344: fixed coloration and made under threshold/flat channels gray
at about 355 to 357: makes it so the opened channel elaborates its mark

THINGS IT HELPS WITH:
speeds up run times and web page load times by reducing channels being ploted


uberpye:
[Here is a sample of the results](https://ldas-jobs.ligo-la.caltech.edu/~jeffrey.bidler/detchar/code_test/testF/). I am unsure what cutoffs the group wants for the "danger" "alert" ext. tags. Currently they are at .6 for danger and .4 for alert.

PR #46, opened 2016-12-12
Outliers

tjma12:
Implements a simple version of a standard deviation cut on the range and h(t) BLRMS channels.

### DIFF
--- a/bin/gwdetchar-slow-correlation
+++ b/bin/gwdetchar-slow-correlation
@@ -64,8 +64,8 @@ parser.add_argument('-r', '--range-channel',
                     help='name of range channel to use')
 parser.add_argument('-R', '--range-frametype',
                     help='frametype for --range-channel')
-parser.add_argument('-O', '--remove-outliers', action='store_true',
-                    help='Remove outliers from BLRMS data')
+parser.add_argument('-O', '--remove-outliers', type=float, default=None,
+                    help='Std. dev. limit for removing outliers')
 
 psig = parser.add_argument_group('Signal processing options')
 psig.add_argument('-b', '--band-pass', type=float, nargs=2, default=None,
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@tjma12, can you describe the logic in the new `remove_outliers` method? I'm trying to work out if there's a way to do the operation with array logic, rather than a `for` loop.

duncanmmacleod:
Also, can you please post a before/after example of what this PR does.

tjma12:
@duncanmmacleod For the quick and ugly first version I replaced any data point that was beyond the sigma threshold with the average of the points on either side of it. I wrote in two edge cases, if the bad point is the first point in the data set, it averages the two points in front of it, if it's the last point in the data set it averages the two points behind it. 

Before: https://ldas-jobs.ligo-wa.caltech.edu/~tjmassin/detchar/slow-correlations/removed-outliers/v2/outliers/
After a 2 sigma cut: https://ldas-jobs.ligo-wa.caltech.edu/~tjmassin/detchar/slow-correlations/removed-outliers/v2/removed_outliers_range_blrms/

The r-values go from an average of about 0.2 to an average of about 0.55 and the scaling on the plots works a bit better without big peaks pushing around the scaling.

duncanmmacleod:
would it work just as well to simply ping the outliers to the 2-sigma value?

```python
mean_ = ts.mean()
threshold = N * ts.std()
ts[ts>mean_+threshold] = mean_ + threshold
ts[ts<mean_-threshold] = mean_ - threshold
```

tjma12:
I don't think so, at least not reliably, it would probably require a lot of tuning to get the right sigma value. Here's an example with the mean and +/- 2 sigma lines outlined, replacing the peaks would still leave a bit of an outlier behind: https://ldas-jobs.ligo.caltech.edu/~thomas.massinger/detchar/smooth_range/RMS.png

PR #50, opened 2017-01-19
Minute trend time correction

uberpye:
clips the ends of the data to fit minute trends

### DIFF
--- a/bin/gwdetchar-overflow
+++ b/bin/gwdetchar-overflow
@@ -40,6 +40,13 @@ from gwpy.utils import gprint
 from gwdetchar import (cds, cli, const, daq, __version__)
 from gwdetchar.io import (datafind, ligolw, html as htmlio)
 
+try:
+    from LDAStools import frameCPP
+except ImportError:
+    readkwargs = {'format': 'gwf'}
+else:
+    readkwargs = {'type': 'adc', 'format': 'gwf.framecpp'}
+
 __author__ = 'TJ Massinger <thomas.massinger@ligo.org>'
 __credits__ = 'Duncan Macleod <duncan.macleod@ligo.org>'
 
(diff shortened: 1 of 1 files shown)


bin/gwdetchar-slow-correlation
```diff
@@ -78,6 +80,16 @@ args = parser.parse_args()
 
 start = args.gpsstart.seconds
 end = args.gpsend.seconds
+
+cstart = start
+cend = end
+
+if (args.trend_type == 'minute') and (0 != (end-start)%60):
+		warn("Clipping start and end times to work with minute trends!")
```
duncanmmacleod: please correct the indentation to use 4 spaces, no tabs please.

duncanmmacleod:
@uberpye this change seems to check whether the total duration is a multiple of 60. If would have though that the check needed to be whether the start and end times are individually multiples of 60? If you can explain to me why that's not the case, I'm happy to merge.

uberpye:
@duncanmmacleod, 
what i think you're suggesting 
`((0 != end%60) or (0 != start%60))`
will restrict end and start to only multiples of 60. For example, if end=122 and start=2, the ends will get clipped to 120 and 60 with a 60 second span of time.
what i wrote 
`(0 != (end-start)%60)`
will restrict end and start such that the time span a multiple of 60. Using the above example, the ends stay where they are at 122 and 2, keeping a 120 second span of time.
This will not bother the code as it won't care about the edges so long as the time in between is a multiple of its stride of 60.

uberpye:
start time, end times, and run time are not integer multiples of 60: https://ldas-jobs.ligo-la.caltech.edu/~jeffrey.bidler/detchar/code_test/time_test2/

going to run one where the start times, end times, and run time are integer multiples of 60 so that will be here soon

PR #59, opened 2017-03-29
Removed deprecated function StripTableName

tjma12:
I was able to successfully run the `gwdetchar-mct` executable with this change in place.


tjma12:
Also very quickly fixed an int that should've been a float in the results of `gwdetchar-mct`, the executable that was using this function

PR #61, opened 2017-03-30
Threshold color fix and command argument

uberpye:
Fixed threshold tab coloration to work like plot ordering

added a comand line option: use -t # or --threshold # (where # is the threshold to plot) to change the threshold

color fix: https://ldas-jobs.ligo-la.caltech.edu/~jeffrey.bidler/detchar/code_test/color-test/
arg addition (threshold set at 0.8): https://ldas-jobs.ligo-la.caltech.edu/~jeffrey.bidler/detchar/code_test/command-test/

### DIFF
--- a/bin/gwdetchar-slow-correlation
+++ b/bin/gwdetchar-slow-correlation
@@ -67,6 +67,8 @@ parser.add_argument('-R', '--range-frametype',
                     help='frametype for --range-channel')
 parser.add_argument('-O', '--remove-outliers', type=float, default=None,
                     help='Std. dev. limit for removing outliers')
+parser.add_argument('-t', '--threshold', type=float, default=0.1, 
+                    help='threshold for making a plot')
 
 psig = parser.add_argument_group('Signal processing options')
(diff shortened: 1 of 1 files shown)


bin/gwdetchar-slow-correlation
```diff
@@ -295,6 +300,7 @@ def process_channel(input_,):
 
 pool = multiprocessing.Pool(nprocplot)
 results = pool.map(process_channel, auxdata.iteritems())
+#results = pool.map(partial(process_channel, thresh=threshold), auxdata.iteritems())
```
uberpye: they are not needed they were a product of me trying to clean up the commits, and i cannot remove yet through git.

duncanmmacleod: it's just 06b0490d1867614d8be3683fa2b14792961441da that is causing problem, I think, you should be able to execute

```
git revert 06b0490d1867614d8be3683fa2b14792961441da
```

to remove it, then push.

uberpye: I removed it by redoing those changes (easy now but I know it will be a bad idea to do that later), and i will use that next time.

PR #63, opened 2017-04-18
Channel diff

andrew-lundgren:
This script could still use a little tweaking, but it's already functional and fast. It reads second-trend frames and finds all channels that appear to be state channels (defined as ones that don't change for 8 seconds). Then it compares the value at two different times chosen by the user, and reports all that change. The intent is to find settings that may have caused the IFO to go from a good to a bad state.

A bit better would be to check whether any channels change during a period (a few seconds for a glitch, could do longer if the transition is more gradual). That would catch things like the CPU usage spiking during the glitch, and not rely on error channels that latch on.

There is one bug showing up:
```
RuntimeWarning: invalid value encountered in subtract
  return a[slice1]-a[slice2]
```
Some channels don't seem to have values that can be interpreted by diff. We'll need to specifically catch this and figure out what's going on.


andrew-lundgren:
I'm not sure why the first merge shows up. That was just me bringing my fork up to date.

bin/gwdetchar-channel-diff
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
```
duncanmmacleod: @andy, please add the coding statement and the standard GPLv3 license header after the hashbang. You can copy directly from any of the other `/bin/gwdetchar-*` executables

bin/gwdetchar-channel-diff
```diff
@@ -0,0 +1,67 @@
...
+    chans = channels[idx:idx+10000]
+    data_dict1=TimeSeriesDict.read(cache1, chans, t1-8, t1,
+                                format='framecpp', type='adc', nproc=8)
+    data_dict2=TimeSeriesDict.read(cache2, chans, t2, t2+1,
+                                format='framecpp', type='adc', nproc=8)
+    for chan in chans:
+        data1=data_dict1[chan].value
+        data2=data_dict2[chan].value
+        # Check for unchanging channel
+        if any(diff(data1) != 0.):
```
duncanmmacleod: it's much more efficient (I think) to use

```python
        if diff(data1).any():
```

I think

bin/gwdetchar-channel-diff
```diff
@@ -0,0 +1,67 @@
...
+            continue
+
+        val1 = data1[-1]
+        val2 = data2[0]
+
+        if val1 == val2:
+            result.append((chan, 'same', val1))
+        else:
+            result.append((chan, val1, val2))
+            print '%s changes from %.4g to %.4g'%(chan,val1,val2)
```
duncanmmacleod: please treat `print` as a function, not a statement, for python3 compatibility, e.g.

```python
            print('%s changes from %.4g to %.4g' % (chan, val1, val2))
```

bin/gwdetchar-channel-diff
```diff
@@ -0,0 +1,67 @@
...
+
+# Write out the results
+with open('changes-%u-%u.txt' % (t1,t2),'w') as fout:
+    for line in result:
+        if type(line[1]) is not str:
```
duncanmmacleod: it's better to use

```python
        if not isinstance(line[1], str):
```

But, this conditional begs the question, why are you recording channels as `'same'`? Why not just skip these?

bin/gwdetchar-channel-diff
```diff
@@ -0,0 +1,67 @@
...
+        val2 = data2[0]
+
+        if val1 == val2:
+            result.append((chan, 'same', val1))
+        else:
+            result.append((chan, val1, val2))
+            print '%s changes from %.4g to %.4g'%(chan,val1,val2)
+
+# Write out the results
+with open('changes-%u-%u.txt' % (t1,t2),'w') as fout:
```
duncanmmacleod: Can you add the output filename as a command-line argument? Or just print to the screen by default and allow the user to pipe manually to a file?

bin/gwdetchar-channel-diff
andrew-lundgren: This is vestigial debugging code. Probably best to split it off into its own thing - a code for finding channels that don't typically change during lock.

bin/gwdetchar-channel-diff
```diff
@@ -0,0 +1,85 @@
...
+    conn = datafind.connect()
+    cache = conn.find_frame_urls(ifo_lett, ifo_lett+'1_T',
+                                start, end, urltype='file')
+    return cache
+
+import argparse
+parser = argparse.ArgumentParser(description='Find state changes.')
+parser.add_argument('-i','--ifo', type=str, help='IFO', required=True)
+parser.add_argument('-s','--t1', type=int, help='First GPS time', required=True)
+parser.add_argument('-e','--t2', type=int, help='Second GPS time', required=True)
```
duncanmmacleod: @andrew-lundgren, is there a reason to use `required` positional arguments, rather than just options? The `gwdetchar.cli` module provides some default options that you can use, e.g.

```python
from gwdetchar import cli
parser = cli.create_parser(description=__doc__)
cli.add_gps_start_stop_arguments(parser)
cli.add_ifo_option(parser)
```

[snippet from [`gwdetchar-software-saturations`](https://github.com/ligovirgo/gwdetchar/blob/d7a356f792e64c7f568f3cbdb9c10b3a29243072/bin/gwdetchar-software-saturations#L257)]

This then allows users to call 

```bash
gwdetchar-sofware-saturations gpsstart gpsend --ifo X1
```

duncanmmacleod:
@andrew-lundgren, I'm still getting to grips with the review process on github.com, please consider the changes regarding use of `argparse` and `gwdetchar.cli`, then I'll be happy.

alurban:
@andrew-lundgren, @duncanmmacleod, are we still interested in merging this?

alurban:
As @duncanmmacleod, @jrsmith02, and I discussed at the March 2019 LVC meeting, we will close this pull request and open another to add a script called `gwdetchar-wrath-of-conlog`.

PR #72, opened 2017-12-21
Add scatter plots for range and blrms

macedo22:
Hi Duncan,

I added to the file code that generates two scatter plots for each auxiliary channel: 1) channel vs strain and 2) channel vs range. When I run this, it generates an HTML page output like so (https://ldas-jobs.ligo-la.caltech.edu/~alexandra.macedo/detchar/O1/L1-slow-correlation-1133906417-7199/), where the two scatter plots have been added below the other plots for each channel.

Disclaimers: With your last commit, c05f107 (https://github.com/ligovirgo/gwdetchar/commit/c05f1070baa9648d9902fdcc4592c61576c3ae2c#diff-c611d65ec72daf5387046f33ba3faed8), and its changes to gwdetchar-slow-correlation, I am unable to run gwdetchar-slow-correlation with the virtual environment install via these instructions that you and Dr. Smith provided (https://wiki.ligo.org/viewauth/DetChar/O1NoiseBreathing). In order to run gwdetchar-slow-correlation at all, it only works if I source detchar's install of gwpysoft and point to my own code with your last commit reverted (meaning, I undid the 2 changes you made to gwdetchar-slow-correlation that were made in commit c05f107). However, it seems you may be working on a new module, nds2, so after I got my code to work with the commit reverted, I replaced the lines to reflect your latest commit so that it is consistent with what you are currently working on. Please refer to the CSUF slack thread for the command that generates the error and the error message, itself, when I try to run gwdetchar-slow-correlation with the virtual environment's install of gwpysoft.

### DIFF
--- a/bin/gwdetchar-slow-correlation
+++ b/bin/gwdetchar-slow-correlation
@@ -305,8 +305,6 @@ def process_channel(input_,):
         # plot scatter plots
         rangeColor = 'red'
         darmblrmsColor = 'blue'
-        #plotHeight = 10
-        #plotWidth = 20
 
         tsCopy = ts.reshape(-1, 1)
         rangetsCopy = rangets.reshape(-1, 1)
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@macedo22, apologies this has lingered for this long, is it still ready to go?

macedo22:
@duncanmmacleod it works fine as-is, but I think I will clean it up a bit first and push a new commit in the near future. I will notify you when I do

duncanmmacleod:
@macedo22, bump.

PR #78, opened 2018-03-01
Reformat model information and add link to full channel list

macedo22:


bin/gwdetchar-lasso-correlation
```diff
@@ -654,6 +654,18 @@ if args.cluster is True:
...
     clusters = pool.map(generate_clusters, enumerate(results[0:len(useful)]))
 
+channelsFile = '%s-CHANNELS-%s.txt'% (args.ifo, gpsstub)
+
+try:
+    numpy.savetxt(channelsFile, channels, fmt="%s")
+except(IOError, IndexError):
+    numpy.savetxt(channelsFile, channels, fmt="%s")
+except RuntimeError as e:
+    if 'latex' in str(e).lower():
```
duncanmmacleod: @macedo22, why would latex get involved here?

macedo22: @duncanmmacleod Not sure. I am not too familiar with Latex and because this exception handling was used for saving all of the plots, I assumed it would go here too. I figured the handling was to account for characters in strings that Latex would get mad at, but it sounds like what you're saying is a Latex exception wouldn't be raised from saving channel names to a text file, as it could for images. I will remove the if/else conditional here and just have 
```except RuntimeError:
       raise```

Comment (duncanmmacleod) [reply bin/gwdetchar-lasso-correlation]: there should be no need for any exception handling here at all, so I recommend you remove all of it, unless you understand a configuration where the file writing should be allowed to fail.

**Comment by macedo22** [comment]
Hi @duncanmmacleod, I removed the error handling you requested as well as a few other things.
1. `process_channel` now only iterates through the channels with a non-zero lasso coefficient
2. The boolean clustering argument was defaulting to `True` before, but it is now fixed and works as expected.
3. Non-zero channels and their lasso coefficients are now stored in a dictionary and accessed by key instead of indexed in an array.

File: bin/gwdetchar-lasso-correlation [hunk 2]
```diff
@@ -99,20 +100,15 @@ parser.add_argument('-J', '--nproc-plot', type=int, default=None,
 parser.add_argument('-o', '--output-dir', default=os.curdir,
                     help='output directory for plots')
 parser.add_argument('-f', '--channel-file', type=os.path.abspath,
-                    help='path for channel file')
+                    help='path for channel file', required=True)
```
duncanmmacleod: @macedo22, I'm not a fan of required optional arguments, since that makes no sense? Can you have a think about changing this to a positional argument if it really is `required`?

bin/gwdetchar-lasso-correlation
```diff
@@ -128,7 +124,7 @@ psig.add_argument('-x', '--filter-padding', type=float, default=3.,
 lsig = parser.add_argument_group('LASSO options')
 lsig.add_argument('-a', '--alpha', default=None, type=float,
                   help='alpha parameter for LASSO fit')
-lsig.add_argument('-C', '--cluster', default=True, type=bool,
+lsig.add_argument('-C', '--cluster', default='True', type=str,
```
duncanmmacleod: I don't understand this change, can you explain?

bin/gwdetchar-lasso-correlation
```diff
@@ -198,37 +215,37 @@ auxdata = TimeSeriesDict.get(
...
-
 for k, ts in auxdata.items():
     flat = ts.value.min() == ts.value.max()
     if flat:
         flatdata[k] = ts
     else:
         gooddata[k] = ts
-
 auxdata = gooddata
+flattab=Table(data = (numpy.asarray(flatdata.keys()),), names = ('Channels',))
```
duncanmmacleod: @macedo22, please use pep8 white-spacing standards throughout.

bin/gwdetchar-lasso-correlation
macedo22: Yes, I can! I can just remove it to keep options open, as you say

bin/gwdetchar-lasso-correlation
macedo22: @duncanmmacleod As I discovered from troubleshooting and reading more online, for some reason, boolean arguments will always be `True` whether you pass in `True` or `False`. There are a few different ways to work around this issue, but I found this the simplest and most readable.

bin/gwdetchar-lasso-correlation
macedo22: @duncanmmacleod will do!

bin/gwdetchar-lasso-correlation
macedo22: I believe it has something to do with the nature of python or parsing. It will treat any string passed as a boolean as `True`

macedo22:
@duncanmmacleod I went ahead and put in your edits. In addition, I have swapped out `LassoCV` as the default process to use to find the best alpha value when none is given. Jeff and Marissa have been experimenting with information criterion, and it seems to be more useful and informative for our purposes. I replaced `LassoCV` with using information criterion to find the best alpha to use if none is given.

bin/gwdetchar-lasso-correlation
duncanmmacleod: in this case you want to use [`action='store_true'`](https://docs.python.org/2/library/argparse.html#action) and just give the argument with a value, e.g.

```python
lsig.add_argument('-C', '--cluster', action='store_true', 
                  help='generate clustered channel plots')
```

the user should then pass just `--cluster` to enable the clustering.

bin/gwdetchar-lasso-correlation
```diff
@@ -675,29 +748,34 @@ page.h2('Parameters')
...
 page.div(class_='model-body')
 
 page.div(class_='model-info')
-page.p('<font size ="4">Model: LASSO<br />Non-zero coefficients: %d<br />Alpha: %g<br />Zero coefficients: %d (%s)</font>' % (numpy.count_nonzero(model.coef_), usedalpha, len(zeroedResults), "<a href= %s target='_blank'>zeroed channel list</a>" % (zeroFile)))
+write_param('Model', 'LASSO')
+write_param('Non-zero coefficients', '%d' % numpy.count_nonzero(model.coef_))
+write_param('Alpha', '%g' % usedalpha)
+write_param('Zero coefficients','%d (%s)' % (len(zeroedtab), "<a href= %s target='_blank'>zeroed channel list</a>" % (zerofile)))
 page.p('<br /><br />%s' % df.to_html(index=True))
 page.div.close()#model-info
```
duncanmmacleod: as per pep8, please leave two spaces before inline comments:

```
page.div.close()  # model-info
```

bin/gwdetchar-lasso-correlation
macedo22: @duncanmmacleod I'm happy to change this, but our overall usage preference has been to always do clustering by default, and only turn it off if in special cases (e.g. when troubleshooting the code overall to fix bugs). If we would like to keep it as doing clustering by default without the user passing anything, is there another way you would like me to rewrite it? If you still feel this new way you proposed is best, I will make the switch.

duncanmmacleod: @macedo22, you can use the opposite logic and provide something like:

```python
lsig.add_argument('-C', '--no-cluster', action='store_true', default=False,
                  help='do not generate clustered channel plots')
```

I believe that one of those options using `action='store_true'` is better than having the user manually pass `true` or `false` on the command-line.

macedo22: @duncanmmacleod Sounds good to me. I swapped this in and tested it, so it should be good to go with this most recent commit `56d11d5`.

macedo22:
Hi @duncanmmacleod, I did an overhaul of the code in an attempt to make it more robust. The changes include:
-the large summary plots at the top are generated before the individual auxiliary channel plots
-instead of using `pyplot` to edit and set features of the plots, the `Figure` objects themselves are used
-we kept getting `TimeoutError`s and `LOCKERROR`s, and the added `config_mpl()` function is meant to reconfigure matplotlib to store each individual plot's tex cache in its own location, instead of them all sharing the same tex cache folder. This is to avoid two or more plots simultaneously trying to acquire lock with the shared tex folder. Since this was added, we are not encountering any of the above errors anymore.

macedo22:
An updated example page output: https://ldas-jobs.ligo-la.caltech.edu/~alexandra.macedo/detchar/LASSO/test/O2/L1/20170301/1172363755-1172375019/

duncanmmacleod:
@macedo22, this looks fine to me, the CI failure is from somewhere else and is related to backwards-incompatible gwpy changes that haven't been addressed in gwdetchar yet, I'll get to it soon.

macedo22:
@duncanmmacleod Great! It seemed like the issue was in `gwdetchar-scattering`, but let me know if there ends up being any unanticipated compatibility issues with `gwdetchar-lasso-correlation` and I can address them on this branch before merge

duncanmmacleod:
@macedo22, I fixed that failure in #87, so if you can rebase your feature branch against the upstream `master` and force push, you should be able to get past that.

macedo22:
@duncanmmacleod Done and done!

PR #92, opened 2018-07-30
This commit along with a minor change to glue (which I madea PR for) …

scottcoughlin2014:
…allow this to be run in py35

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -27,7 +27,10 @@ import re
 import sys
 import ast
 import warnings
-from StringIO import StringIO
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
 
 import numpy
 from numpy import fft as npfft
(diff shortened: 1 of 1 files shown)


alurban:
@scottcoughlin2014, does this PR require waiting for your other PR in `glue` to be merged, or is it ready right now?

bin/gwdetchar-omega
```diff
@@ -27,7 +27,10 @@ import re
 import sys
 import ast
 import warnings
-from StringIO import StringIO
+try:
+    from StringIO import StringIO
+except ImportError:
+    from io import StringIO
```
duncanmmacleod: @scottcoughlin2014 (@alurban) can we swap the ordering of this to make it python3-first?

scottcoughlin2014: Done

scottcoughlin2014:
@alurban Sure this can be merged without the `glue` merge. all it means is that it will still not work in python 3 until that is merged but not that this will all of a sudden stop working or whatever.

alurban:
Ok, I will merge. Thanks @scottcoughlin2014 for putting this together and @duncanmmacleod for reviewing

PR #99, opened 2018-08-01
Updated travis configuration

duncanmmacleod:
This PR cleans up the travis configuration a little:

- use `python -m` to guarantee the right interpreter
- use `$PIP_FLAGS` consistently
- removed all references to custom `lal` builds
- removed all `apt` package installs (we don't need them any more)


duncanmmacleod:
@alurban, this might conflict with #98, so we should rebase this once that has been merged and rerun the CI.

PR #115, opened 2018-09-04
Modified gwdetchar-omega to use Helvetica Neue font (not Lato)

duncanmmacleod:
This PR modifies the `gwdetchar.omega` module such that `gwdetchar-omega` now produces HTML that uses the Helvetica Neue font, and not Lato. IMO this is more accessible, since its a little easier to read.

I also made some font modifications (make font smaller, a bit).

See the example [here](https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwdetchar/testing/1126259462.4-font/).

### DIFF
--- a/gwdetchar/omega/html.py
+++ b/gwdetchar/omega/html.py
@@ -86,7 +86,6 @@ body {
 		margin-bottom: 120px;
 		min-height: 100%;
 		font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
-		font-weight: 300;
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@alurban, yes, there is a job failure because gwdetchar is incompatible with gwpy-0.12.0. I can roll back to 0.11.0 and rerun, if you would like.

alurban:
@duncanmmacleod, if that's not too much trouble, that would be great.

PR #116, opened 2018-09-05
Build CSS/JS files as part of the package build

duncanmmacleod:
This PR modifies the package build to compile and minify the CSS and javascript files necessary for `gwdetchar-omega`. This has the following consequences:

- this moves the CSS/JS for `gwdetchar-omega` outside of the python modules, this should make those blocks easier to maintain
- adds two new build dependencies: `jsmin` and `libsass`

In order to achieve this the `write_static_files` method has been rebuilt to copy files on disk into the output directory as needed, rather than writing new files.

cc: @alurban

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ Portfile
 /.eggs/
 /.cache/
 
+# static files build
+/gwdetchar/_static
+
 # Installer logs
 pip-log.txt
 pip-delete-this-directory.txt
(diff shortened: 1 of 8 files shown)


duncanmmacleod:
@alurban, apologies for the delay: https://ldas-jobs.ligo.caltech.edu/~duncan.macleod/gwdetchar/testing/static-build/1126259462.4/.

alurban:
LGTM

PR #121, opened 2018-09-13
Add scikit-learn as a dependency

macedo22:
`scikit-learn` has been added to `install_requirements` in `setup.py` and to `requirements.txt`


alurban:
@macedo22, oh no, I didn't realize this was here when I merged #185 ... sorry about that!

Issue #90, opened 2018-07-30
gwdetchar-omega should use bootstrap-ligo for CSS helpers


duncanmmacleod:
The new `gwdetchar-omega` utility defines a lot of its own CSS. It could make use the [duncanmmacleod/bootstrap-ligo](https://github.com/duncanmmacleod/bootstrap-ligo) repo, which is used by `gwsumm` for colouring and other CSS (and JS) helpers).

GWSumm pulls this in as a git submodule, which is then automatically built (`sass` -> `css`) during the `setup.py build`, see  https://github.com/gwpy/gwsumm/blob/5ac705ba7f13f78ebd81995044692be176b4b0c2/setup.py#L103-L182.

PR #123, opened 2018-09-14
Misc. formatting improvements for gwdetchar-omega

### LINKED ISSUE #122: Misc. gwdetchar-omega formatting issues
* In `gwdetchar-omega`, #120 inadvertently removed interferometer context highlighting from the toggle buttons that allow users to switch between different plot types. It should be straightforward to re-introduce this using bootstrap button groups, where each button in the group is a dropdown.

* The button groups should also be right-aligned within their parent class, and items in the menu (e.g. "raw", "whitened", and "autoscaled" under "Q-transform view") should not be in bold.

* For each channel, it may make more sense to put the table of properties ("Properties of the most significant time-frequency tile") from left to right so that it spans the full width of the parent, so that more plots can be added in the bottom row.

* The very top banner of the output should also not have bold text.

alurban:
@duncanmmacleod, this PR contains a few formatting enhancements for `gwdetchar-omega` output pages, including:

* Restoring IFO-specific context highlighting in plot toggle buttons
* Expand `container` width and adapt its size for different screens
* Reorganizing the channel information display so that the channel name is on its own line, plot toggle buttons are flushed right on the same line as the loudest-tile table (which is now horizontal), and up to three plots are shown on a single line beneath the table
* De-bolden section headings (which will need to be done more gracefully once #90 is addressed)
* Update the output page after every channel has been analyzed, as well as after every channel block, and once again at the very end

Examples may be found here (requires `LIGO.ORG` credentials):

IFO | Scan time | Results
-- | -- | --
Hanford | 1187020580.44 |[results](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/H1_1187020580.44/)
Livingston | 1172489783.070 | [results](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/L1_1172489783.07/)
GEO | 1186970418.0 | [results](https://ldas-jobs.ligo.caltech.edu/~alexander.urban/for_borja/pyomega/G1_1186970418.0/)

This fixes #122.

### DIFF
--- a/gwdetchar/omega/html.py
+++ b/gwdetchar/omega/html.py
@@ -688,7 +688,40 @@ def write_block(block, context, tableclass='table table-condensed table-hover '
             continue
         page.li(class_='list-group-item')
         page.div(class_="container")
+
+        page.div(class_='row', style='margin-top:6pt; width:98%')
+
+        # channel name
+        page.div(class_='col-md-7')
         page.h4(cis_link(channel.name))
+        page.div.close()  # col-md-7
+
+        # plot toggle buttons
+        page.div(class_='col-md-5')
+        page.div(class_='btn-group', role='group')
+        for ptitle, pclass, ptypes in [
+            ('Timeseries', 'timeseries', ('raw', 'highpassed', 'whitened')),
+            ('Q-transform', 'qscan', ('raw', 'whitened', 'autoscaled')),
+            ('Eventgram', 'eventgram', ('raw', 'whitened', 'autoscaled')),
+        ]:
+            _id = 'btnGroup{0}{1}'.format(pclass.title(), i)
+            page.div(class_='btn-group', role='group')
+            page.button(id_=_id, type='button',
+                        class_='btn btn-%s dropdown-toggle' % context,
+                        style='opacity:0.8;', **{'data-toggle': 'dropdown'})
+            page.add('{0} view <span class="caret"></span>'.format(ptitle))
+            page.button.close()
+            page.ul(class_='dropdown-menu', role='menu',
+                    **{'aria-labelledby': _id})
+            for ptype in ptypes:
+                page.li(toggle_link('{0}_{1}'.format(pclass, ptype), channel,
+                                    channel.pranges))
+            page.ul.close()  # dropdown-menu
+            page.div.close()  # btn-group
+        page.div.close()  # btn-group
+        page.div.close()  # col-md-5
+
+        page.div.close()  # row
 
         page.div(class_='row')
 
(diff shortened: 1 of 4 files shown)


gwdetchar/omega/html.py
```diff
@@ -393,7 +391,7 @@ def toggle_link(plottype, channel, pranges):
     chanstring = channel.name.replace('-', '_').replace(':', '-')
     captions = [p.caption for p in channel.plots[plottype]]
     return markup.oneliner.a(
-        '<b>%s</b>' % text, class_='dropdown-item',
+        '%s' % text, class_='dropdown-item',
```
duncanmmacleod: @alurban, can you replace `'%s' % text` with just `text` (since `text` is already a `str`). There's a few more of those below as well.

gwdetchar/omega/html.py
```diff
@@ -592,7 +590,7 @@ def write_summary(
     """
     utc = tconvert(gpstime)
     page = markup.page()
-    page.h2(header)
+    page.h2(header, style='font-weight:normal;')
```
duncanmmacleod: it makes sense to me to keep the style declarations to the scss file(s), where I think they will be easier to manage in the long term. can you have a look at doing that?

alurban: These style choices will eventually be replaced by `bootstrap-ligo` when we get around to doing that, to be consistent with the summary pages. But, to your point, that will be easier if we stick them in the scss files now, so I'll address that shortly.

alurban:
@duncanmmacleod, I've pushed a few commits that address your review comments. I also reorganized the layout for each channel so that the table of loudest-tile properties is horizontal at the top, aligned with the toggle buttons, and a row of three plots is saved at the bottom. The examples for H1 and L1 linked to above have been updated with these changes.

share/sass/gwdetchar-omega.scss
```diff
@@ -27,6 +27,43 @@ body {
...
 		min-height: 100%;
 		font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;
+		font-weight: 300;
+}
+
+h1 {
+		font-weight: normal;
+}
+
+h2 {
```
duncanmmacleod: @alurban, you probably want to simplify this:

```sass
h1, h2, h3, h4, h5, h6 {
		font-weight: normal;
}
```
or define a custom class (e.g. backporting from bootstrap v4):

```sass
.font-weight-normal {
		font-weight: normal;
}
```

alurban: Aha! I didn't know you could do that, I'll try again shortly.

alurban: @duncanmmacleod, thanks for the suggestion, I have made this change and verified that works with the Hanford example (linked to above).

PR #126, opened 2018-09-17
Reduce default false alarm rate threshold to 1e-10 Hz

### LINKED ISSUE #125: Reduce the false alarm rate threshold in omega scans
In `gwdetchar-omega`, there is an option `--far-threshold` which applies a threshold on the "false alarm rate" (FAR) for each channel. Statistically, this means any channel that produces events more often than expected from white Gaussian noise gets removed from the analysis. Functionally, this allows us to downselect the channels that show up on the final output page.

The current default FAR threshold is 0.001 / second, which in many cases still allows hundreds of channels. This can be whittled down substantially by reducing the default FAR threshold to 1e-10 / second (or about one event per 300 years).

alurban:
Note, users wanting to use a different threshold can still specify that using the command line option `--far-threshold`.

This fixes #125.

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -63,7 +63,7 @@ parser.add_argument('-f', '--config-file', action='append', default=None,
                     help='path to configuration file to use, can be given '
                          'multiple times (files read in order), default: '
                          'None')
-parser.add_argument('-t', '--far-threshold', type=float, default=1e-3,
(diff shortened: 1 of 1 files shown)


alurban:
The Travis CI build failure reported here is in a unit test for `sngl_burst`, which this PR doesn't touch. Because the PR is so minimal I'm going to go ahead and merge it.

PR #140, opened 2018-09-25
Hard-code omega scan plots to always use seconds on the time axis

### LINKED ISSUE #139: In gwdetchar-omega, switch to hard-coded time axis units
Per @duncanmmacleod's suggestion: if you want to fix to using a [seconds] scale, you should use

```python
ax.set_xscale('seconds', epoch=gps)
```

alurban:
This fixes #139.

### DIFF
--- a/gwdetchar/omega/plot.py
+++ b/gwdetchar/omega/plot.py
@@ -65,12 +65,8 @@ def omega_plot(series, gps, span, channel, output, colormap='viridis',
                            figsize=figsize)
     ax = plot.gca()
     # set time axis units
-    ax.set_epoch(gps)
-    ax.set_xscale('auto-gps')
-    if span <= 1.:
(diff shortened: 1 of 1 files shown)


alurban:
@areeda, if it helps your review, below is an example plot made using GWPy in a way similar to the changes introduced in this PR:

```python
gps = 1187008882
from gwpy.timeseries import TimeSeries
import matplotlib as mpl; mpl.use('Agg')
data = TimeSeries.get('L1:GDS-CALIB_STRAIN', gps-.5, gps+.5)
plot = data.plot()
ax = plot.gca()
ax.set_xscale('seconds', epoch=gps)
plot.show()
```

test

This is with the cluster install of GWPy 0.11.0 at LLO pcdev1. Note the x-axis is correctly labeled with the right units.

alurban:
thanks @areeda! I'll merge.

PR #142, opened 2018-09-25
Adjust the size of Q-transform spectrograms

### LINKED ISSUE #141: Adjust the size and resolution of Q-transform spectrograms
In `gwdetchar-omega`, the size of each Q-transform spectrogram can be reduced by a factor of 4 (or more!) by using the `outseg` keyword argument to `TimeSeries.q_transform`. 

At the same time, the frequency resolution of these spectrograms should be increased so that the final plots are less choppy at low frequency. The resolution could be different for each channel and chosen based on e.g. the FFT length used to construct the whitening filter, or the lower-frequency cutoff used in the Q transform. (This will wind up increasing the size of the spectrogram as a tradeoff, although it's still a net win if one chooses, say, `flow / 20` instead of the currently chosen `flow / 5`.)

alurban:
This PR includes two changes to `gwdetchar-omega`:

* Use the `outseg` keyword argument to `TimeSeries.q_transform()` to reduce the length of the time axis of Q-transform spectrograms to the duration of the longest plot, rather than the full length of data analyzed. For 16k channels, which are typically configured to read in 64 seconds of data but plot at most 16 seconds, this reduces the total size by a factor of 4. (The reduction factor will be larger for slower channels.)
* Increase the frequency resolution of Q-transform spectrograms to `flow / 20`, where `flow` is the lowest frequency analyzed for a given channel. This replaces the previous value of `flow / 5`, which was insufficient and led to choppiness in the plots at low frequency.

For 16k channels, these two changes effectively cancel each other out in terms of total number of array elements. However, slower channels which require analyzing more data will see a factor > 1 reduction in array size.

This fixes #141.

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -33,6 +33,7 @@ use('agg')  # nopep8
 from gwpy.utils import gprint
 from gwpy.time import tconvert
 from gwpy.table import EventTable
+from gwpy.segments import Segment
 from gwpy.timeseries import TimeSeriesDict
 from gwpy.detector import (Channel, ChannelList)
 from gwpy.signal.qtransform import QTiling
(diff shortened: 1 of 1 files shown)


alurban:
@areeda, here is an example Q-transform spectrogram with the new frequency resolution:

omega_qscan

To my eye, this looks quite smooth at low frequency.

PR #152, opened 2018-09-29
Reduce the data duration read in from frames

### LINKED ISSUE #151: Omega scans should grab only as much data as is requested
To prevent filter transients using the GWPy whitener prior to 0.12, `gwdetchar-omega` is hard-coded to read in 512 seconds (plus padding) of data for every channel. Now that 0.12+ is formally required, we should be reading in only as much data as is requested by the user. This is set using the `duration` config file argument.

alurban:
Now that we require gwpy-0.12+, we should only read in as much data per channel as is requested by the user. This fixes #151.


alurban:
thanks @duncanmmacleod

PR #155, opened 2018-10-01
Support omega scan sub-blocks

### LINKED ISSUE #146: Allow channels in the same block to have different Omega scan properties
In `gwdetchar-omega`, we should require contextual channel blocks to allow channels with different properties to appear in the same block. For example, not all SUS channels have the same sample rate, but we may eventually find it convenient to group them together. For another example, some users may want to have `GDS-CALIB_STRAIN` and `CAL-DELTAL_EXTERNAL_DQ` in the same block for comparison, but that is not currently possible because these channels do not have the same frametype.

It should be relatively simple to implement something like a `parent` argument in INI blocks, such that child blocks inherit properties from their parents and override a subset of them. In the final output page, channels in the parent and child would appear in the same contextual block.

alurban:
This PR makes two substantive changes:

* Add support for a `parent` Omega scan INI kwarg, which allows contextual channel blocks with different analysis attributes to be linked to each other so that they are analyzed separately, but appear in one block together on the output page.
* Ensure that the list of analyzed channels is built from the bottom up, rather than from the top down

[**Here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/L1_170817_withparent/) is an example output page which places `L1:GDS-CALIB_STRAIN` and `L1:CAL-DELTAL_EXTERNAL_DQ` in the same contextual block on the output page (requires `LIGO.ORG` credentials). These channels have different frame types, so I used the `parent` argument to configure them to appear together.

This fixes #146.

### DIFF
--- a/gwdetchar/omega/html.py
+++ b/gwdetchar/omega/html.py
@@ -525,19 +525,15 @@ def write_summary_table(blocks, base=os.path.curdir):
 
     Parameters
     ----------
-    blocks : `list` of `OmegaChannelList`
+    blocks : `dict` of `OmegaChannel`
         the channel blocks scanned in the analysis
     base : `str`
         the path for the `<base>` tag to link in the `<head>`
     """
     # record summary data for each channel
     channel, time, freq, Q, energy, snr = ([], [], [], [], [], [])
-    for block in blocks:
-        for chan in block.channels:
-            try:  # record channels only if they were analyzed
-                chan.energy
-            except AttributeError:
-                continue
+    for block in blocks.values():
+        for chan in block['channels']:
             channel.append(chan.name)
             time.append(chan.t)
             freq.append(chan.f)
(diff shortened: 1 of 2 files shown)


alurban:
@duncanmmacleod, after merging #158 I rebased this branch against master. Somewhat surprisingly, there were no merge conflicts, so this PR is ready for review.

gwdetchar/omega/html.py
```diff
@@ -856,7 +848,7 @@ def write_qscan_page(blocks, context):
     page = markup.page()
     page.add(write_toc(blocks))
     page.h2('Channel details')
-    for block in blocks:
+    for key, block in blocks.items():
```
duncanmmacleod: shouldn't this just be `for block in blocks.values()`?

alurban: Yes it should -- I think in the course of testing I made several changes, but forgot to go back and switch this from `items()` to `values()`. I did this in a couple of other places too, which is fixed in the latest commit.

PR #158, opened 2018-10-02
Allow Omega scan input to be read from a local cache

### LINKED ISSUE #153: Allow frame caches in gwdetchar-omega
`gwdetchar-omega` should allow contextual channel blocks to come from different sources. In particular, users should be able to point to a local file cache if they want to run omega scans over non-published data and display them on the same page. This can be done by refactoring frame I/O into a separate function, and introducing an optional `source` INI argument that points to the file cache, which by default is `None` or overwritten by `frametype`.

alurban:
This PR refactors data I/O into a separate function, `get_data`, which allows data to be read from  NDS/frames (by calling `TimeSeriesDict.get()`) or from a cache (using `TimeSeriesDict.read()`). It also introduces support for a `source` INI argument, which allows users to point to a file cache when reading in individual channel blocks. (If the `source` argument is given, it will override `frametype`.)

[**Here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/L1_cache_test/) is an example comparing online Livingston data to some local frames I made for calibration testing way back in early O2. The online data were read from C00 HOFT frames, while the test data were read out of a local file cache. 

This fixes #153.

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -145,6 +145,7 @@ class OmegaChannelList(object):
         self.duration = int(params.get('duration', 32))
         self.fftlength = int(params.get('fftlength', 2))
         self.resample = int(params.get('resample', 0))
+        self.source = params.get('source', None)
         self.frametype = params.get('frametype', None)
         chans = params.get('channels', None).split('\n')
         self.channels = [OmegaChannel(c, self.name, **params) for c in chans]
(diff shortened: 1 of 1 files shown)


bin/gwdetchar-omega
```diff
@@ -153,6 +154,51 @@ class OmegaChannelList(object):
...
+    end = time + duration/2. + pad
+    # read from cache if requested
+    if source:
+        data = TimeSeriesDict.read(source, channels, start=start, end=end,
+                                   nproc=nproc)
+    # read from frames or NDS
+    else:
+        data = TimeSeriesDict.get(channels, start, end, frametype=frametype,
+                                  nproc=nproc, verbose=verbose)
+    return data
```
duncanmmacleod: @alurban, I would refactor this a little, mainly for cleanliness as:

```python
    # read from cache if requested
    if source:
        return TimeSeriesDict.read(source, channels, start=start, end=end,
                                   nproc=nproc)
    # read from frames or NDS
    return TimeSeriesDict.get(channels, start, end, frametype=frametype,
                              nproc=nproc, verbose=verbose)
```

alurban: @duncanmmacleod, good suggestion! I made this change and double-checked it by updating the test page.

PR #160, opened 2018-10-04
Refresh omega scans until they finish

### LINKED ISSUE #159: Force-reload omega scan output until the scan finishes
It would be nice to have a way to force-reload omega scan pages until the scan finishes. This can be accomplished using `<meta>` HTML tags to refresh from the server side once per minute, e.g. `<meta http-equiv="refresh" content="60">`.

alurban:
cc @areeda 

This PR introduces light functionality that force-refreshes Omega scan output pages until the scan finishes running.

Example output may be found [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/L1_cache_test/). While the scan was running this page auto-refreshed once every 60 seconds, and has stopped auto-refreshing since then, as designed.

This fixes #159.

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -92,6 +92,7 @@ cp.read(config_files)
 htmlv = {
     'title': '%s Qscan | %s' % (ifo, gps),
     'config': config_files,
+    'refresh': True,
 }
 
 # set output directory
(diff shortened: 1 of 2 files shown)


alurban:
Thanks @areeda!

PR #161, opened 2018-10-05
Patched gwdetchar-overflow

tjma12:
Glue doesn't seem to like using the `glue.lal.Cache.sieve()` method on individual segments, so I modified the script to use a single element `SegmentList` instead. Tested at CIT and the executable worked.


alurban:
LGTM, thanks @tjma12

PR #164, opened 2018-10-08
fixed x axis label for second trends

uberpye:
There was a small error in the plot labels when you use second trends.

### DIFF
--- a/bin/gwdetchar-lasso-correlation
+++ b/bin/gwdetchar-lasso-correlation
@@ -184,10 +184,17 @@ args = parser.parse_args()
 start = int(args.gpsstart)
 end = int(args.gpsend)
 pad = args.filter_padding
-auto_xlabel = ('Time [hours] from '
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
@uberpye, the `auto_xlabel` variable you are modifying here is never actually used in the `gwdetchar-lasso-correlation` script, AFAICT. Can you confirm this?

uberpye:
> 
> 
> @uberpye, the `auto_xlabel` variable you are modifying here is never actually used in the `gwdetchar-lasso-correlation` script, AFAICT. Can you confirm this?

@duncanmmacleod, looks like the reason for `auto_xlabel`'s existance got fixed by commit [680fa4d](https://github.com/gwdetchar/gwdetchar/commit/bfee8c7d763f1f46cfdb7c08248b733b1680fa4d) so we don't need my commit or that variable anymore. woops.

PR #174, opened 2018-10-17
Enabled using nds2 in gwdetchar-overflow

duncanmmacleod:
This PR modifies the `gwdetchar-overflow` tool to allow using NDS2 for data access, as required to use this tool at KAGRA.

This is WIP because I have not tested whether the code _still works_ when reading GWF files directly. I will do that shortly and post a comment with the result.

### DIFF
--- a/bin/gwdetchar-overflow
+++ b/bin/gwdetchar-overflow
@@ -49,6 +49,11 @@ except ImportError:
 else:
     readkwargs = {'type': 'adc', 'format': 'gwf.framecpp'}
 
+SITE_MAP = {
+    'K1': 'KAGRA',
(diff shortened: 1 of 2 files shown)


alurban:
@duncanmmacleod, can you post an examples with (say) KAGRA and one of the LIGO detectors?

duncanmmacleod:
@alurban, KAGRA data are not available outside KAGRA, so I cannot post an example. [Here](https://ldas-jobs.ligo-la.caltech.edu/~duncan.macleod/gwdetchar/testing/overflow-nds/) is an example of the updated script running on LIGO _without using NDS_ as proof that this change is backwards compatible.

PR #176, opened 2018-10-18
Allow users to require state flags to be active to process omega scans

### LINKED ISSUE #144: In gwdetchar-omega: prevent analyses outside a given segment
In `gwdetchar-omega` there is a config argument called `state-flag` which currently does nothing. This argument should be used to declare a segment safe for analysis, i.e. the argument should prevent `gwdetchar-omega` from analyzing any times when the given segment is not active.

alurban:
cc @duncanmmacleod, @areeda, @tjma12 

This PR allows users to require state flags to be active before processing omega scans. For example, if a channel block is configured with the following lines:

```
duration = 64
state-flag = L1:DMT-GRD_ISC_LOCK_NOMINAL:1
```

then the flag `L1:DMT-GRD_ISC_LOCK_NOMINAL:1` must be active during the 64 seconds (plus padding) surrounding the given GPS time. If no channels wind up being analyzed, a null page is written with the message `No significant channels found during active analysis segments`.

[**Here**](https://ldas-jobs.ligo-la.caltech.edu/%7Eaurban/pyomega-test/L1_unlocked/) is an example run during a time when `L1:DMT-GRD_ISC_LOCK_NOMINAL:1` is not active, and [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/L1_170817_withparent/) is an example when it is (to verify that the change is backward compatible). Both require `LIGO.ORG` credentials.

This fixes #144.

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -34,9 +34,9 @@ use('agg')  # nopep8
 from gwpy.utils import gprint
 from gwpy.time import to_gps
 from gwpy.table import EventTable
-from gwpy.segments import Segment
 from gwpy.timeseries import TimeSeriesDict
 from gwpy.detector import (Channel, ChannelList)
+from gwpy.segments import (Segment, DataQualityFlag)
 from gwpy.signal.qtransform import QTiling
 
 from gwdetchar import (cli, __version__)
@@ -59,6 +59,9 @@ parser.add_argument('-f', '--config-file', action='append', default=None,
                     help='path to configuration file to use, can be given '
                          'multiple times (files read in order), default: '
                          'None')
+parser.add_argument('-s', '--ignore-state-flags', action='store_true',
+                    default=False, help='ignore state flag definitions in '
+                                        'the configuration, default: False')
 parser.add_argument('-t', '--far-threshold', type=float, default=1e-10,
                     help='White noise false alarm rate threshold for '
                          'processing channels; default: %(default)s Hz')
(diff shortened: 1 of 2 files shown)


bin/gwdetchar-omega
```diff
@@ -158,6 +159,40 @@ class OmegaChannelList(object):
...
+    # set GPS start and end time
+    start = time - duration/2. - pad
+    end = time + duration/2. + pad
+    inseg = Segment(start, end)
+    # query for state segments
+    aseg = DataQualityFlag.query(flag, start, end).active
+    # check that state flag is active during the entire analysis
+    if (not aseg.intersects_segment(inseg)) or (len(aseg) > 1):
+        return False
+    elif abs(aseg[0]) < abs(inseg):
```
duncanmmacleod: @alurban, I think the `len(aseg) > 1` and `abs(aseg[0]) < abs(inseg)` checks are redundant, so we should only test one.

duncanmmacleod:
@alurban, this looks fine, but it would be nice (IMO) to add a command-line option to allow users to ignore the state flags if they want to. This will help, for example, when the segment database is down.

bin/gwdetchar-omega
alurban: @duncanmmacleod, aha, it seems you're right:

```
[======   ===]
[     =======]
```

In both examples the code wouldn't run because the flag isn't active during the entire segment of interest, and checking that `abs(aseg[0]) < abs(inseg)` is sufficient to establish this. This is reflected in the latest commit.

alurban:
@duncanmmacleod, good idea, I've added a command line option `--ignore-state-flags` that does this.

alurban:
thanks @duncanmmacleod!

PR #178, opened 2018-10-18
Cleaning up omega scan command line options

### LINKED ISSUE #177: Omega scan verbose argument does nothing
The `--verbose` argument in `gwdetchar-omega` currently has no effect because the default is `'False'` (a string), when it *should* be `False` (a `bool`). As a result, the output `--verbose` is supposed to activate is just always on.

Elsewhere, the `--ifo` option should be set using `gwdetchar.cli.add_ifo_option(parser)`.

alurban:
This PR ensures that the command line options for `gwdetchar-omega` are functional with the following changes:

* Set the default for `--verbose` to `False` (rather than the string `'False'`, which led to verbose output always being turned on)
* Use `gwdetchar.cli.add_ifo_option()` to add the optional IFO argument
* Define the `gpstime` argument so that its type is `gwpy.time.to_gps`, allowing users to pass either GPS seconds or datesting formats such as `'2018-10-18 01:07:56'`
* Have `gwdetchar.cli.add_nproc_option()` print its default in the help string

[**Here**](https://ldas-jobs.ligo-la.caltech.edu/%7Eaurban/pyomega-test/L1_170817_withparent/) is an example run with these changes.

This fixes #177.

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -32,7 +32,7 @@ from matplotlib import (use, rcParams)
 use('agg')  # nopep8
 
 from gwpy.utils import gprint
-from gwpy.time import tconvert
+from gwpy.time import to_gps
 from gwpy.table import EventTable
 from gwpy.segments import Segment
 from gwpy.timeseries import TimeSeriesDict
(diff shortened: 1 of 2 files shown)


alurban:
thanks @areeda!

PR #189, opened 2018-11-09
Cross-correlate Omega scan channels

### LINKED ISSUE #95: Add a table of ranked channels to the output of gwdetchar-omega
Add a table of the top N channels ranked by some figure of merit. This should include a drop-down menu from which users may select different figures of merit to rank by, such as SNR or signal energy. 

To assist users in finding witness channels, there should also be a similarity metric, preferably based on the timeseries. Perhaps a max cross-correlation within the search window. (Note, signals in various witnesses may be time-shifted due to a time-of-flight delay.)

### LINKED ISSUE #221: Refactor gwdetchar-omega
Now that `gwdetchar-omega` is stable and robust, we should refactor parts of it into callable functions so that the code is more modular. This will benefit downstream consumers, as well as simplify anticipated new features, such as #189.

alurban:
cc @duncanmmacleod, @siddharth101, @areeda, @scottcoughlin2014

This PR takes advantage of a new feature introduced in gwpy-0.13.0 to provide a similarity metric for Omega scans. The new feature, `TimeSeries.correlate`, allows users to designate one channel as a primary, whiten it, then cross-correlate with every other whitened channel in the scan. As @siddharth101 has shown, this can help quickly identify witness channels for glitches seen in h(t), assuming there is no frequency conversion.

Also added is a table of the top 5 channels ranked by correlation with the primary, with the following features:

* Table caption includes a link to a fancybox plot of the whitened primary timeseries
* The first column is a link to the channel view
* The table is sorted in descending order by correlation amplitude
* The time delay (in ms) between aux and h(t) is also shown, enforcing a window of [-100 ms, +100 ms] around the scan time
* The first channel listed is always the primary
* Channels whose matched-filter output has corrupted normalization are not shown in the table, but are still shown in the full scan

Finally, to simplify the workflow (and allow users direct access to intermediate data products), this PR also refactors `gwdetchar-omega` into a new module with core utilities:

* Add a new module `gwdetchar.omega.core` which is inherited by `__init__`
* Remove `gwdetchar.omega.plot.omega_plot` and replace it with a utility to plot all data products from an omega scan called `write_qscan_plots`
* Rename `gwdetchar/omega/scan.py` as `gwdetchar/omega/matlab.py` to prevent confusion with the new `gwdetchar.omega.scan` utility (this file will be removed soon anyway)

Below are links to full Omega scans run with this these changes:

IFO | Date | Glitch type | Link
----|----|----|----
H1 | 2019-01-03 | Scattering | [**Scans**](https://ldas-jobs.ligo-wa.caltech.edu/~aurban/wdq/20190203/)
L1 | 2019-01-23 | Whistles, scattering | [**Scans**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/wdq/20190216/)

For more information, please see gwpy/gwpy#905. The new features are **not** a significant memory hog, according to condor logs of memory usage vs. time.

This fixes #95 and fixes #221.


bin/gwdetchar-omega
```diff
@@ -433,6 +447,50 @@ html.write_qscan_page(ifo, gps, analyzed, **htmlv)
...
+        plot.omega_plot(hoft, gps, primary.length, primary.channel.name,
+                        'plots/primary.png', ylabel='Whitened Amplitude',
+                        figsize=[8, 5])
+        hoft = hoft.crop(
+            gps - primary.length/2, gps + primary.length/2).taper()
+        del data
+    # prepare HTML output
+    htmlv['correlated'] = active
+    htmlv['primary'] = primary.channel.name
+
```
duncanmmacleod: @alurban, can we move some or all of this block into a function, to make it a bit more modular?

duncanmmacleod: @alurban, any thoughts on this?

alurban: @duncanmmacleod, sorry, I thought I had already responded. Yes, I think that refactoring much of this code block into a function is a very good idea that is on my to-do list. (In fact, much of the [**other refactoring work**](https://github.com/gwpy/gwpy/pull/984) I've been doing has been a preamble to this.)

alurban: @duncanmmacleod, we discussed this yesterday, and decided it will save effort to get this feature in now and save refactoring the bulk of this code into a function for a future pull request.

alurban: @duncanmmacleod, I've pushed a commit that _heavily_ refactors `gwdetchar-omega` into a new module, `gwdetchar.omega.core`, that is loaded from `gwdetchar/omega/__init__.py`. It also renames `gwdetchar/omega/scan.py` --> `gwdetchar/omega/matlab.py`, to avoid clobbering a new function I introduced called `scan()` that does all the heavy lifting.

PR #199, opened 2018-11-18
Refactor get_data and check_flag into gwdetchar.io.datafind

alurban:
This PR refactors the functions `check_flag` and `get_data`, which were originally defined in `gwdetchar-omega`, into `gwdetchar.io.datafind` so that they can be used for more general purposes. It also adjusts `get_data()` so that that function directly calls `TimeSeriesDict.read()` after constructing a list of `Cache` objects using `gwdetchar.io.datafind.find_frames()`.

cc @areeda, @duncanmmacleod


alurban:
cc @areeda, @duncanmmacleod 

In the course of putting this PR together I tried three separate different frame reading methods:

* Read in one-channel-at-a-time using gwpy's `TimeSeries.read()` method with the keyword argument `format='gwf.lalframe'`, which @duncanmmacleod suggested may reduce the memory overhead
* Read in one-channel-at-a-time using gwpy's `TimeSeries.read()` method, but without passing an argument for `format` (so that it defaults to `'gwf'`)
* Read in all channels for a given contextual block up-front using `TimeSeriesDict.read()`, with no argument for `format` (again, it defaults to `'gwf'`)

I ran Omega scans over the [**ten loudest Omicron triggers**](https://ldas-jobs.ligo-la.caltech.edu/~detchar/summary/day/20181117/lock/glitches/) from 2018-11-17 using each frame I/O method, keeping the run configuration fixed, and compared total memory usage and time to completion. Full Omega scans are here:

Frame I/O Method | Output Directory
------ | -------
`TimeSeries.read()` + `format='gwf.lalframe'` | [**Link**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/wdq/20181117/)
`TimeSeries.read()` + default `format` | [**Link**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/wdq/20181117_framecpp/)
`TimeSeriesDict.read()` + default `format` | [**Link**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/wdq/20181117_tsdict/)

It turns out that these methods have almost no impact whatsoever on total memory usage (see Fig. 1), while the `TimeSeriesDict.read()` method tends to finish in the least amount of time (see Fig. 2). Presumably this is because the same amount of data is being read in either way, but `TimeSeriesDict.read()` only checks the frame table of contents once per contextual channel block. Based on these results, this is the frame I/O method I'm going with.

image

image

PR #202, opened 2018-11-20
Refactor Omega scan config utilities into gwdetchar.omega.config

alurban:
cc @areeda, @duncanmmacleod 

This PR refactors some Omega scan utilities which were originally defined in `bin/gwdetchar-omega` into `gwdetchar.omega.config`. This includes the following changes:

* Move `OmegaChannel` and `OmegaChannelList` into `gwdetchar.omega.config`, rearranging them a little in anticipation of #189. Note, this does not impact the Matlab scans because those use entirely separate classes defined in `gwdetchar.omega.scan` that happen to be called the same thing.
* Move `get_fancyplots` into `gwdetchar.omega.config`
* Write a helper function to grab default config files given IFO and GPS time
* Add a method `OmegaConfigParser.get_channel_blocks()` to populate an ordered dict of contextual channel blocks for analysis

Example output under these changes is available [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/Network_170817/).

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -48,7 +48,6 @@ from __future__ import division
 
 import os
 import sys
-import ast
 import numpy
 from scipy.signal import butter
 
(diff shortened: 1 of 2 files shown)


alurban:
Thanks @areeda

PR #205, opened 2018-11-29
Resample Q-transform spectrograms along the time axis

alurban:
This PR introduces resampling of Q-transform spectrograms along the time axis to save time and memory. The resampling is done in `gwdetchar.omega.plot()` by slicing along the time axis.

This PR also removes explicit deletion of objects in the main loop in `gwdetchar-omega`, because these have no impact on the overall memory usage.

cc @areeda, @duncanmmacleod

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -320,13 +320,11 @@ for block in blocks.values():
         except UnboundLocalError:
             if args.verbose:
                 gprint('Channel is misbehaved, removing it from the analysis')
-            del series, hpseries, wseries, asd
             continue
         if table.Z < table.engthresh and not c.always_plot:
             if args.verbose:
                 gprint('Channel not significant at white noise false alarm '
                        'rate %s Hz' % far)
-            del series, hpseries, wseries, asd, table
             continue
         Q = table.q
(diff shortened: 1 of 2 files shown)


alurban:
@areeda, below are some statistics for memory usage and total runtime with the proposed changes. I did two batch-mode runs through condor, with each run covering the ten loudest L1 Omicron triggers from November 17. I chose this day because there's a nice mix of relatively quiet and loud times.

The first plot shows total memory usage as reported in the condor log file, the second shows total time to completion as measured in system hours.

image

image

As you might expected, slicing the longer spectrograms along their time axes does reduce memory consumption by about 10-20% on average, and speeds up the runs a little too.

I've got [a pull request open in gwpy](https://github.com/gwpy/gwpy/pull/983) which log-samples along the frequency axis, and this should reduce memory consumption and runtime even more. That functionality will be ready for the next release, likely 0.13.0.

alurban:
@areeda yeah that's right, I forgot to mention the buffer deletion above. I'll edit that now. I removed those because it turned out that they had zero impact on memory consumption, and a future PR will refactor all that into a function anyway, so the garbage collector will take care of it.

PR #208, opened 2018-12-04
Log-normalize the colorbar for autoscaled spectral plots

### LINKED ISSUE #207: Auto-lognorm in autoscaled Q-scan spectrograms
@areeda, a while ago you and I were talking about color autoscaling in Q-scan spectrograms like this one:

image

The issue is that the peak energy is so high compared to the median (~1e6) that the background is a washed-out sea of dark blue. What do you think of automatically defaulting to a log-norm (i.e., a log-scaled colorbar) in such cases? I'm thinking maybe if the peak energy is larger than 100 we can switch to a log-norm.

alurban:
cc @areeda 

This PR changes `gwdetchar.omega.plot` so that spectral plots (spectrograms and eventgrams) with an autoscaled colorbar are log-normalized with `vmin=0.5`.

This fixes #207.

### DIFF
--- a/gwdetchar/omega/plot.py
+++ b/gwdetchar/omega/plot.py
@@ -80,19 +80,22 @@ def omega_plot(series, gps, span, channel, output, colormap='viridis',
     if (qscan or eventgram):
         ax.set_yscale('log')
         title = '%s at %.3f with $Q$ of %.1f' % (channel, gps, Q)
+        cmap = cm.get_cmap(colormap)
+        ax.set_facecolor(cmap(0))
+        if clim is not None:  # linear colorbar with the requested limits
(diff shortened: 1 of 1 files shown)


alurban:
@areeda, I have example output pages for [**a Virgo-only scan**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/V1_1186741861.0/) and a [**full IFO network scan**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/Network_170817/) at the time of GW170817. To see the changes, use the bootstrap buttons to toggle to "autoscaled spectrogram" and "autoscaled eventgram."

PR #223, opened 2019-02-18
Fixed some labelling issues when not using TeX with matplotlib

duncanmmacleod:
This PR fixes some matplotlib labelling issues when _not_ using TeX.


duncanmmacleod:
I think having the conditionals is good, so that if we choose again to move away from TeX (fingers crossed), we don't have to update this again.

PR #230, opened 2019-02-21
Deprecate support for Matlab omega scans

alurban:
This pull request removes support for Matlab omega scans entirely from `gwdetchar`:

* Remove `gwdetchar/omega/matlab.py`
* Remove `bin/wdq`
* Rename `bin/wdq-batch --> bin/gwdetchar-omega-batch`
* Remove options for `wdq` and `wpipeline` from `gwdetchar-omega-batch`
* Raise a `DeprecationWarning` for `wdq-batch`

cc @duncanmmacleod


duncanmmacleod:
@alurban, thanks for this; I'm ok with removing `wdq`, but I think we should retain `wdq-batch` for a little bit, since that tool has not had a deprecation warning associated with it, so its removal might come as a bit of a shock.

alurban:
@duncanmmacleod, that's a very good point, I've renamed it back to `wdq-batch` and added a `DeprecationWarning`.

PR #247, opened 2019-03-07
Allow users to opt out of TeX

duncanmmacleod:
This PR patches `gwpy.plot.get_gwpy_tex_settings` to allow users to opt-out of using TeX for plotting, using the `GWPY_USETEX` environment variable supported by gwpy.

### DIFF
--- a/gwdetchar/plot.py
+++ b/gwdetchar/plot.py
@@ -21,6 +21,7 @@
 
 from matplotlib import rcParams
 from gwpy.plot.tex import (has_tex, MACROS as GWPY_TEX_MACROS)
+from gwpy.utils.env import bool_env
 
 __author__ = 'Alex Urban <alexander.urban@ligo.org>'
 __credits__ = 'Dan Hoak <daniel.hoak@ligo.org>, ' \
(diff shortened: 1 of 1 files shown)


alurban:
@duncanmmacleod, I'm torn on this one. On the one hand, it would definitely be nice to allow users to bypass TeX rendering if they wish. But on the other, the motivation for setting up `gwdetchar.plot.get_gwpy_tex_settings` in the first place was to enforce that omega scans _not_ use TeX in our production runs. If `GWPY_USETEX` is set, then the upshot is that _all_ plot rendering in `gwdetchar` would use TeX.

It seems the preferred option would be to allow users to toggle TeX usage for omega scans independently of the rest of `gwdetchar`. What do you think?

duncanmmacleod:
@alurban, I don't understand. The current default (before this change) is to _always_ use TeX, with no choice given to the user. My change is to enable using `GWPY_USETEX=no` to allow people to manually deselect TeX.

I am _not_ suggesting we, as service managers, set `GWPY_USETEX` in production, but I (as a user) will set `GWPY_USETEX=no` mainly because TeX is slow and not required for speedy development.

PR #248, opened 2019-03-07
Set LaTeX params at runtime, rather than at the module level

alurban:
There was a bug in omega scan plots where rendering was done with LaTeX because TeX params were set in `gwdetchar.plot` and then imported by `gwdetchar.omega.html`. This PR corrects that by (reverting to) handling all TeX settings in scripts at runtime, rather than at the module level.

cc @duncanmmacleod

### DIFF
--- a/bin/gwdetchar-mct
+++ b/bin/gwdetchar-mct
@@ -25,9 +25,6 @@ import numpy
 import os
 import sys
 
-from matplotlib import (use, rcParams)
-use('agg')
-
 import gwdatafind
 
 from glue.ligolw.utils import write_fileobj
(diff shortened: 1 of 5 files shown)


bin/gwdetchar-mct
```diff
@@ -25,9 +25,6 @@ import numpy
 import os
 import sys
 
-from matplotlib import (use, rcParams)
-use('agg')
-
```
duncanmmacleod: Is this removal just because `gwdetchar-mct` doesn't actually plot anything?

alurban: @duncanmmacleod, yes, that's right, I figured it was best to remove the redundancy.

PR #249, opened 2019-03-07
Read LASSO primary channel from the requested frametype

alurban:
This fixes a frame reading bug in `gwdetchar-lasso-correlation` discovered by @marissawalker and Guillermo Valdes.


alurban:
@marissawalker signed off on slack, so I'll go ahead and merge.

PR #253, opened 2019-03-08
Address feature requests for O3 scattering

### LINKED ISSUE #252: Feature requests for O3 scattering pages
This request comes from @jrsmith02:

To O3-ize the [**scattering pages**](https://ldas-jobs.ligo-la.caltech.edu/~detchar/summary/day/20190307/detchar/scattering/) I think we should:

* Add the new Z mirrors upper stage longitudinal channels (just like the other channels, but Z mirrors), `{IFO}:SUS-ZM1_M{1,2}_DAMP_L_IN1_DQ`
* Add some 30s padding near the start and end of locks to reduce false alarms (if not already there)
* Hide the segments as .txt files because currently when we open an accordion segment you have to scroll forever and ever to find the graph.

alurban:
This addresses feature requests for O3 scattering pages, in particular:

* 30 seconds [default] of padding after the start and before the end of each analysis segment, configurable by the user
* For each collapsible panel, move plots above active scattering segment tables

Example output is available [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/scattering/day/20190302/), compare with the [**production version**](https://ldas-jobs.ligo-la.caltech.edu/~detchar/scattering/day/20190302/index.html) (both require `LIGO.ORG` credentials).

This fixes #252. Note, the LIGO observatories' ZM1 and ZM2 mirrors sit on single-stage tip-tilt suspensions, so they don't have an `M2` stage (and `M1` is already analyzed by default).

cc @jrsmith02, @andrew-lundgren, @duncanmmacleod

### DIFF
--- a/bin/gwdetchar-scattering
+++ b/bin/gwdetchar-scattering
@@ -135,10 +135,15 @@ if args.state_flag:
     state = DataQualityFlag.query(args.state_flag, int(args.gpsstart),
                                   int(args.gpsend),
                                   url=const.DEFAULT_SEGMENT_SERVER)
+    padding = args.segment_start_pad + args.segment_end_pad
     for i, seg in enumerate(state.active):
-        if abs(seg) > (args.segment_start_pad + args.segment_end_pad):
+        if abs(seg) > padding:
             state.active[i] = type(seg)(
                 seg[0]+args.segment_start_pad, seg[1]-args.segment_end_pad)
+        else:
+            if args.verbose:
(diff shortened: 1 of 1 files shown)


alurban:
Thanks @duncanmmacleod, your input is always most welcome :)

@jrsmith02, can you have a look at the test output posted above and let me know if you're satisfied?

jrsmith02:
Yeah that looks great! All three things I asked for (with one already having been done). Thank you!

PR #259, opened 2019-03-11
Moved finalize_static_urls function upstream to io.html

duncanmmacleod:
This PR moves the `gwdetchar.omega.html.finalize_static_urls` function upstream into `gwdetchar.io.html`, mainly so that it can be used by other tools without muddying the namespaces.

### DIFF
--- a/gwdetchar/omega/html.py
+++ b/gwdetchar/omega/html.py
@@ -22,13 +22,9 @@
 from __future__ import division
 
 import os
-import sys
 import numpy
 from functools import wraps
-from shutil import copyfile
 from collections import OrderedDict
-
-from six.moves.urllib.parse import urlparse
 
 from pkg_resources import resource_filename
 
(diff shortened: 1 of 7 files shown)


gwdetchar/omega/html.py
```diff
@@ -381,8 +303,7 @@ def close_page(page, target, about=None, date=None):
         `~datetime.datetime.now`
     """
     page.div.close()  # container
-    page.add(write_footer(about=about, date=date, class_=True,
-                          linkstyle='color:#eee;'))
+    page.add(io_html.write_footer(about=about, date=date))
```
alurban: @duncanmmacleod, some of the unit tests are failing because this line needs to retain the `class_` and `linkstyle` kwargs:

```suggestion
    page.add(io_html.write_footer(about=about, date=date, class_=True,
                                  linkstyle='color:#eee;'))
```

`class_=True` will render the footer as a `footer` class, rather than a simple `div`, while the `linkstyle` is `'color:#000;'` by default (which is no good for omega scan footers).

gwdetchar/omega/tests/test_html.py
```diff
@@ -280,28 +280,6 @@ def test_fancy_plot():
     assert test.caption is 'test.png'
 
 
-def test_finalize_static_urls(tmpdir):
-    static = os.path.join(str(tmpdir), 'static')
```
alurban: @duncanmmacleod, I think this line may be the reason for the other build failures. The functions that fail all call `finalize_static_urls`, so they all expect to find a `static` directory, but don't. This is probably my fault for not explicitly including this line in the other unit tests to begin with, apologies for that.

Can you add this in to the other unit tests?

alurban: Actually, I take that back, `finalize_static_urls` should automatically create this directory. Hmm...

PR #264, opened 2019-03-12
Added package_list and package_table functions

duncanmmacleod:
This PR adds some utility functions to `gwdetchar.io.html` to scape the environment from `pip`/`conda` and render as an html table, and added this to the about page for `gwdetchar-omega`.

Example output is [here](https://ldas-jobs.ligo-la.caltech.edu/~duncan.macleod/gwdetchar/testing/package-table/GW170817/about/) (LIGO.ORG).

### DIFF
--- a/gwdetchar/io/html.py
+++ b/gwdetchar/io/html.py
@@ -19,10 +19,13 @@
 """Utilties for HTML output
 """
 
+import json
 import os
 import sys
 import datetime
+import subprocess
 from getpass import getuser
+from operator import itemgetter
 from shutil import copyfile
 try:
     from pathlib2 import Path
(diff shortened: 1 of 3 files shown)


gwdetchar/io/html.py
```diff
@@ -475,3 +478,77 @@ def write_footer(about=None, date=None, class_=False,
...
+    page.tbody()
+    for pkg in sorted(pkgs, key=itemgetter("name")):
+        page.tr()
+        for col in cols:
+            page.td(pkg[col.lower()])
+        page.tr.close()
+    page.tbody.close()
+    page.table.close()
+
+    return page
```
alurban: @duncanmmacleod, I've been making a point to return markup objects like so:

```suggestion
    return page()
```

This allows you to append html more simply, as `page.add(package_table())` rather than `page.add(str(package_table())`. Can you make that change?

gwdetchar/omega/html.py
```diff
@@ -842,4 +842,6 @@ def write_about_page(configfiles):
         with open(configfile, 'r') as fobj:
             inifile = fobj.read()
         page.add(htmlio.render_code(inifile, 'ini'))
+    # add environment listing
+    page.add(str(htmlio.package_table()))
```
alurban: @duncanmmacleod, apropos the comment above, this line can be done more simply:

```suggestion
    page.add(htmlio.package_table())
```

duncanmmacleod: As discussed privately, I have not changed this, since this way we don't rely on the upstream function returning a `str` explicitly. If we decide to change the return upstream in the future, we don't have to change this again.

PR #265, opened 2019-03-12
Removed unused imports

duncanmmacleod:
This PR removes unused imports from all modules. We should set up flake8 to run as part of the CI to catch these automatically. GWpy does this already, so we can probably steal from there.


alurban:
LGTM, and using `flake8` is a good idea, we should definitely do that in the future

PR #271, opened 2019-03-19
Fixed bug reading limit channels in software-saturation tool

duncanmmacleod:
This PR fixes a bug determining the relevant channels in the `gwdetchar-software-saturations` tools, when the cache doesn't contain an entry at the 'start' time.

### DIFF
--- a/gwdetchar/saturation.py
+++ b/gwdetchar/saturation.py
@@ -30,6 +30,7 @@ import numpy
 from astropy.units import Quantity
 from multiprocessing import (cpu_count, Pool)
 
+from gwpy.io.cache import file_segment
 from gwpy.timeseries import StateTimeSeries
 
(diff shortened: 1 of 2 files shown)


gwdetchar/saturation.py
```diff
@@ -186,7 +186,8 @@ def is_saturated(channel, cache, start=None, end=None, indicator='LIMEN',
             channels[i] = c[:-6]
     # check limit if set
     indicators = ['{}_{}'.format(c, indicator) for c in channels]
-    data = get_data(indicators, start, start+1, source=cache, nproc=nproc)
+    gps = file_segment(cache[0])[0]
```
alurban: @duncanmmacleod, I think this is missing an import statement for the file_segment utility, otherwise looks good to me.

PR #273, opened 2019-03-21
Introducing gwdetchar-conlog

### LINKED ISSUE #272: create gwdetchar-wrath-of-conlog
We use the tool gwpy-wrath-of-conlog written by Andy Lundgren to do analyses such as [this one](https://alog.ligo-la.caltech.edu/aLOG/index.php?callRep=42555). This tool seems appropriate for gwdetchar. So we could:

1) bring 'gwpy-wrath-of-conlog' into gwdetchar, with appropriate name change

To help with this, slide 4 of [this pdf](https://alog.ligo-la.caltech.edu/aLOG/uploads/42555_20190101112112_noise-jumps-wrath-conlog.pdf) contains a link to the location of a running version of the code and a use case. We note that `gwdetchar` has a utility that will handle data discovery and automatically ignore all channels that don’t exist in frames for the entire duration of the time requested, so that should improve the code.

2) In addition, we should create a standard channel list (or list filter) based on the lab's use of conlog [here](https://alog.ligo-wa.caltech.edu/aLOG/index.php?callRep=47279), but still allow the option to run on all channels or give a channel list.

3) We should consider also including scripts to combine multiple outputs of gwd-conlog, like [this one](https://ldas-jobs.ligo-la.caltech.edu/~jrsmith/detchar/preO3/wrath-of-conlog/all_three.py).

alurban:
This PR introduces a new command-line tool (`gwdetchar-conlog`) that identifies trends showing significant change between any two GPS times. Given a start time, stop time, IFO, and frametype, the script first identifies all channels that exist at both times, then reads frame data and prints channels whose values have changed.

`gwdetchar-conlog` also includes the following options:
* `--nproc`: number of parallel cores to use while reading data frames
* `--output`: optional filename to record output in, defaults to `changes.txt`
* `--channels`: optional list of regular expression patterns against which to require matching channel names, reducing the amount of frame reading and thus memory consumption and runtime

This fixes #272.

cc @duncanmmacleod, @jrsmith02, @andrew-lundgren, @tjma12


bin/gwdetchar-conlog
```diff
@@ -0,0 +1,107 @@
+#!/home/aurban/opt/gwdetchar-dev/bin/python
```
duncanmmacleod:
```suggestion
#!/usr/bin/env python
```

bin/gwdetchar-conlog
```diff
@@ -0,0 +1,107 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with GW DetChar.  If not, see <http://www.gnu.org/licenses/>.
+
+"""Check whether a list of 16Hz EPICS records changed between two reference
+times
+"""
+
+from __future__ import (division, print_function)
```
duncanmmacleod:
```suggestion
```

bin/gwdetchar-conlog
```diff
@@ -0,0 +1,107 @@
...
+
+import re
+import numpy
+import warnings
```
duncanmmacleod:
```suggestion
```

bin/gwdetchar-conlog
```diff
@@ -0,0 +1,107 @@
...
+    # flat channels
+    if any(numpy.diff(xoft1) != 0):
+        continue
+
+    # dynamic channels
+    value1 = xoft1[-1]
+    value2 = xoft2[0]
+    if value1 == value2:
+        continue
+    else:
```
duncanmmacleod: This `else` can be removed:

```python
    if value1 == value2:
        continue
    messages.append('{} changes from {:.4g} to {:.4g}'.format(
            channel, value1, value2))
    print(messages[-1])

alurban:
@duncanmmacleod, thanks for taking a look, I've addressed your review comments. I've also tested the code and saved the output [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/conlog/changes-1229317908-1229317928.csv) (there is also a log file [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/conlog/output.txt)).

jrsmith02:
I love it. Likely we will tweak it as we use it. But LGTM

PR #275, opened 2019-03-22
Enhancements to gwdetchar-conlog

alurban:
This PR enhances the functionality of `gwdetchar-conlog` based on conversations with @andrew-lundgren:

* Allow a pre-baked channel list to be passed to the `--channels` argument, and move regex patterns to a new argument called `--search`
* If `--channels` is passed, process the union of these with the set of channels available in frames
* Save the difference between reference values of channels that get processed
* Print the table of channel diffs directly to `stdout`
* Misc. formatting and flake8 fixes

cc @duncanmmacleod, @andrew-lundgren, @jrsmith02

### DIFF
--- a/bin/gwdetchar-conlog
+++ b/bin/gwdetchar-conlog
@@ -22,6 +22,7 @@ times
 """
 
 import re
+import numpy
 
 import gwdatafind
 from gwpy.table import Table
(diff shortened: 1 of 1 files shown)


alurban:
@duncanmmacleod, that's a suggestion I agree with, so I went ahead and implemented that (and a few minor formatting tweaks when printing to `stdout`). I've tested the code and saved the output [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/conlog/output.txt) (requires `LIGO.ORG` credentials).

PR #279, opened 2019-03-26
Use a colo(u)rful version of python's logging utility

### LINKED ISSUE #181: Use Python's logging facility in gwdetchar tools
@duncanmmacleod, @areeda: I propose that, before packaging `gwdetchar`, we switch to using [Python's `logging` facility](https://docs.python.org/3/library/logging.html) to print to stdout/stderr. The primary motivation, at least to me, is that this makes it easier to set the logging level (particularly with verbose output).

alurban:
This PR uses a combination of python's built-in `logging` module, and a third-party plugin called `coloredlogs`, to provide nicely formatted colo(u)rful logging to `stdout` for all `gwdetchar` scripts.

This fixes #181.

### DIFF
--- a/bin/gwdetchar-conlog
+++ b/bin/gwdetchar-conlog
@@ -22,6 +22,8 @@ times
 """
 
 import re
+import os.path
+
 import numpy
 
 import gwdatafind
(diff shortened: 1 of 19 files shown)


alurban:
@duncanmmacleod, here's what the new logging looks like when I run an omega scan:

Screen Shot 2019-03-26 at 16 09 39

and here it is for conlog:

Screen Shot 2019-03-26 at 16 17 08

PR #283, opened 2019-03-27
Crop off unrequested data from minute trends

alurban:
There is a bug upstream of `gwdetchar.io.datafind.get_data` where minute trends occasionally return more data than is requested by `gwdatafind`, which currently breaks `gwdetchar-lasso-correlation` when the primary is read from raw frame files, as those are not affected by this bug (but minute trends are). This PR provides a temporary workaround by cropping off the unwanted data whenever this occurs.

See [**here**](https://ldas-jobs.ligo-wa.caltech.edu/~aurban/lasso/for_josh/) for test output.

cc @duncanmmacleod, @andrew-lundgren, @jrsmith02


andrew-lundgren:
Looks fine, but the data-getting functions also need to be debugged because they're returning an hour more data than requested in some cases.

PR #284, opened 2019-03-28
Condor checkpointing for omega scans

### LINKED ISSUE #200: Implement condor checkpointing for Omega scans
In `gwdetchar-omega`, add support for condor checkpointing so that jobs which fall over (or get held) do not start all the way back at the beginning upon being relaunched. Since the new code already writes a summary of channels processed to disk (in three formats), this might be a convenient way to keep track of where a given scan left off.

alurban:
This PR implements condor checkpointing for `gwdetchar-omega`:

* Take advantage of the fact that, if a given time has already been analyzed, there will exist a `summary.csv` file containing information about channels that were processed above that run's false alarm rate threshold
* Use that file to load summary information about previously analyzed channels
* Get as far as you can with this information, throwing useful errors if anything breaks

This has the added benefit that channels which _weren't_ above the previous run's FAR threshold will still be analyzed, meaning this functionality supports re-running analyses to allow higher FARs.

This fixes #200.

cc @duncanmmacleod, @areeda

### DIFF
--- a/bin/gwdetchar-omega
+++ b/bin/gwdetchar-omega
@@ -177,16 +177,12 @@ if os.path.exists(summary) and not args.disable_checkpoint:
         os.path.abspath(summary)))
     record = Table.read(summary)
     completed = list(record['Channel'])
-    if not args.disable_correlation:
-        try:
-            check_corr = record[0]['Standard Deviation']
-        except KeyError:
-            raise KeyError(
-                'Cross-correlation is not available from this record, '
-                'consider running without correlation or starting from '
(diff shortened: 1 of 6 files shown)


alurban:
@duncanmmacleod, [**Here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/pyomega-test/Network_170817/) is an example of a checkpointed scan, with the command line output shown below.

Screen Shot 2019-03-28 at 12 36 57

alurban:
Thanks @duncanmmacleod, I've restored that hyphen and will now merge 😄

PR #286, opened 2019-03-29
Fix frequency axis limits in scattering plots

alurban:
This PR fixes an issue @siddharth101 pointed out, where plots of fringe frequency as a function of time should display everything down to 0 Hz, while trigger plots should take their limits from `args.fmin`.

Example output may be found [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/scattering/day/20190104/) (requires `LIGO.ORG` credentials).

cc @siddharth101, @duncanmmacleod


alurban:
Thank you @siddharth101

PR #291, opened 2019-04-01
Allow scattering code to trigger omega scans

### LINKED ISSUE #287: Omega scans for gwdetchar-scattering
Now that `gwdetchar-scattering` provides a record of Omicron triggers that fall within active scattering segments, we have everything we would need to also provide the following:

* List other properties of each trigger, such as peak frequency, duration, bandwidth, and SNR
* Allow the user to automatically launch omega scans for at least a subset of these triggers and visualize their output on the scattering page, similar to the way [**`hveto`**](https://github.com/gwdetchar/hveto/blob/master/bin/hveto#L101-L104) does

cc @siddharth101

alurban:
This PR adds an option `--omega-scans`, whose argument is an integer, to `gwdetchar-scattering`. When `--omega-scans N` is nonzero, a random sample of `N` triggers that fall within active scattering segments is chosen to omega scan, with jobs launched in batch mode and displayed on the scattering page.

This PR also refactors the heavy lifting in `gwdetchar-omega-batch` into a new submodule, `gwdetchar.omega.batch`, including unit tests.

Test output is available [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/scattering/day/20190331/) (requires `LIGO.ORG` credentials).

This fixes #287.

cc @duncanmmacleod, @siddharth101

### DIFF
--- a/bin/gwdetchar-scattering
+++ b/bin/gwdetchar-scattering
@@ -547,7 +547,7 @@ nscans = min(args.omega_scans, len(table))
 if nscans > 0:
     # launch scans
     ind = random.sample(range(0, len(table)), nscans)
-    omegatimes = [t for t in table['trigger_time'][ind]]
+    omegatimes = [str(t) for t in table['trigger_time'][ind]]
     logger.debug('Collected {} event times to omega scan: {}'.format(
         nscans, ', '.join(omegatimes)))
     logger.info('Creating workflow for omega scans')
--- a/bin/gwdetchar-scattering
+++ b/bin/gwdetchar-scattering
@@ -546,7 +546,7 @@ nscans = min(args.omega_scans, len(table))
 if nscans > 0:
     # launch scans
     ind = random.sample(range(0, len(table)), nscans)
-    omegatimes = [str(t) for t in table['trigger_time'][ind]]
+    omegatimes = list(table['trigger_time'][ind])
     logger.debug('Collected {} event times to omega scan: {}'.format(
         nscans, ', '.join(omegatimes)))
     logger.info('Creating workflow for omega scans')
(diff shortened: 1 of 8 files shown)


siddharth101:
Thanks a lot @alurban . This looks pretty good.

bin/gwdetchar-scattering
```diff
@@ -532,6 +541,67 @@ print(table)
...
+        '--ifo', args.ifo,
+        '--condor-timeout', str(4),
+        '--submit'
+    ]
+    logger.debug("$ {}".format(" ".join(batch)))
+    proc = subprocess.Popen(batch)
+    out, err = proc.communicate()
+    if proc.returncode:
+        raise subprocess.CalledProcessError(
+            proc.returncode, ' '.join(batch))
```
duncanmmacleod: can you just use `subprocess.check_call` here?

alurban: @duncanmmacleod, aye, I've implemented this just now

alurban:
@duncanmmacleod, that's probably a good idea, since there are now three tools (`gwdetchar-omega-batch`, `gwdetchar-scattering`, and `hveto`) that launch omega scans to condor. I'll have a go at that after I get back from lunch.

alurban:
@duncanmmacleod, at the risk of scope creep, I've refactored the heavy-lifting parts of `gwdetchar-omega-batch` into a new module called `gwdetchar.omega.batch` and used this to launch omega scans from `gwdetchar-scattering`. This change includes unit tests for the new submodule.

Once it's released, we will also be able to use this to clean up similar functionality in [**`hveto`**](https://github.com/gwdetchar/hveto/blob/master/bin/hveto#L813-L839). I particularly like this because it removes a call to `subprocess`, it gives me the warm-and-fuzzies when we can do everything in memory.

gwdetchar/io/tests/test_html.py
```diff
@@ -256,6 +291,14 @@ def test_write_flag_html_with_plots(tmpdir):
     shutil.rmtree(str(tmpdir), ignore_errors=True)
 
 
+def test_scaffold_omega_scans():
+    times = [1126259462]
+    channel = 'X1:STRAIN'
+    page = html.scaffold_omega_scans(times, channel)
+    print(page)
```
duncanmmacleod: Can this be removed?

alurban: thanks @duncanmmacleod, I've removed it, this was left over from when I was writing the unit test and hadn't woken up enough yet

PR #293, opened 2019-04-03
Add protections for the get_data utility

alurban:
This PR fixes a bug in the `get_data` utility by adding protections against nonexistent frame files when attempting to build a frame file cache, and falling back to `TimeSeries{Dict}.get` rather than `fetch` when that fails. Unit tests are also updated to reflect these changes.

Other changes include:
* Remove redundant `obs` keyword argument from `get_data`, and infer the observatory from the `frametype` (when applicable) using a regular expression
* Remove references to `obs` from a few scripts
* Simplify a couple of unit tests
* Update docstrings
* Misc. typo fix in `gwdetchar-conlog`

cc @duncanmmacleod

### DIFF
--- a/gwdetchar/io/datafind.py
+++ b/gwdetchar/io/datafind.py
@@ -20,6 +20,7 @@
 """
 
 import warnings
+from six.moves.urllib.error import HTTPError
 
 import gwdatafind
 
@@ -153,7 +154,7 @@ def get_data(channel, start, end, obs=None, frametype=None, source=None,
 
     **kwargs : `dict`, optional
         additional keyword arguments to `~gwpy.timeseries.TimeSeries.read`
-        or `~gwpy.timeseries.TimeSeries.fetch`
+        or `~gwpy.timeseries.TimeSeries.get`
 
     Returns
     -------
@@ -168,7 +169,7 @@ def get_data(channel, start, end, obs=None, frametype=None, source=None,
 
     See Also
     --------
-    gwpy.timeseries.TimeSeries.fetch
+    gwpy.timeseries.TimeSeries.get
         for the underlying method to read from an NDS server
     gwpy.timeseries.TimeSeries.read
         for the underlying method to read from a local file cache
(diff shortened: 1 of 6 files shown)


gwdetchar/io/datafind.py
```diff
@@ -178,23 +172,19 @@ def get_data(channel, start, end, obs=None, frametype=None, source=None,
...
-        source = gwdatafind.find_urls(obs, frametype, start, end)
     # read from frames or NDS
-    if source:
+    if source is not None:
         if isinstance(channel, (list, tuple)):
             channel = remove_missing_channels(channel, source)
         return series_class.read(
             source, channel, start=start, end=end, nproc=nproc,
             verbose=verbose, **kwargs)
     elif isinstance(channel, str):
```
duncanmmacleod: it occurs to me that this should probably be

```python
        elif not isinstance(channel, (list, tuple))
```

to match the condition above that declares `series_class`

alurban: thanks @duncanmmacleod, I've changed this in the latest commit. I also restored the logic that attempts to load a cache using `gwdatafind` because I remembered that it's _much_ simpler that way, but provided a lightweight wrapper around `gwdatafind.find_urls` to add some protection against non-existent frame files.

alurban:
@duncanmmacleod, I just saw your longer comment, sorry about that. Response below:

I prefer to have `gwdetchar.io.datafind.get_data` attempt to create its own file cache before falling back to `TimeSeries.get`, mainly because it then becomes possible to sieve the file cache against requested channels that actually exist in frame files. This makes `gwdetchar` tools extremely robust to changes in channel names, which is very common between observing runs.

For example, `gwdetchar-scattering` is written to analyze optical squeezer suspensions by default, but it will still run successfully (with default options) over O1 or O2 data when those squeezer optics didn't exist. Likewise for `gwdetchar-omega`, whose default channel list configuration changed several times just in the first few months of 2019.

gwdetchar/io/datafind.py
```diff
@@ -178,23 +179,22 @@ def get_data(channel, start, end, obs=None, frametype=None, source=None,
...
+        source = source or gwdatafind.find_urls(
+            obs or frametype[0], frametype, start, end)
         if isinstance(channel, (list, tuple)):
             channel = remove_missing_channels(channel, source)
         return series_class.read(
             source, channel, start=start, end=end, nproc=nproc,
             verbose=verbose, **kwargs)
-    elif isinstance(channel, str):
-        return series_class.fetch(
+    except (HTTPError, TypeError):  # frame files not found
```
duncanmmacleod: Can you explain when an `HTTPError` would be raised?

alurban: @duncanmmacleod, the `HTTPError` is raised whenever `gwdatafind.find_urls` can't locate frame files (you get a `400 Bad Request`).

alurban: I think that if you don't pass `ops` and the first letter of `frametype` is not a meaningful observatory code (e.g. for SenseMon frames), this will also lead to `HTTPError`. The `TypeError` case comes when you pass neither `frametype` nor `source`. In all these cases, the idea is that it would then fall back to `TimeSeries.get` with no protection against missing channels in frame files.

alurban: @duncanmmacleod, I stand corrected, if you attempt to read SenseMon frames with `frametype='SenseMonitor_hoft_L1_M'` and `obs=None` then the result is an `IORegistryError: Format could not be identified.` Passing `frametype=None` falls back to `TimeSeries.get`, and passing both `frametype` and `obs` successfully uses `TimeSeries.read`.

alurban:
@duncanmmacleod, I've fixed this so that the `obs` kwarg is no longer needed, the observatory is now inferred from `frametype` (when it's passed) using a regular expression. Also, whenever neither `frametype` nor `source` are passed, the utility will fall back to `TimeSeries.get`.

I've also updated the docstring to [hopefully] make it clearer to the user which arguments do what, and when.

PR #295, opened 2019-04-04
HTML table constructors from downstream

### LINKED ISSUE #288: Expand HTML utils for downstream consumers
For downstream consumers, particularly `hveto` and `gwsumm`, we should consider expanding the utilities in `gwdetchar.io.html`:

* Move `write_about_page` out of `gwdetchar.omega.html` and into `gwdetchar.io.html`
* Gracefully handle the case when more than one config file is given, i.e. list the configs as [**expandable panels**](https://ldas-jobs.ligo-la.caltech.edu/~detchar/summary/day/20190331/about/index.html) the way that `gwsumm` does (example requires `LIGO.ORG` credentials)
* Move the [**`navbar`**](https://github.com/gwpy/gwsumm/blob/2dff9ccbdd97140d5f7240abd807efbc40b1d937/gwsumm/html/bootstrap.py#L91-L159) utility out of `gwsumm` and into `gwdetchar.io.html`
* Provide navbars to each output page (why not?)

cc @duncanmmacleod

alurban:
This PR brings in a few HTML table constructors from downstream:

* Move `gwsumm.html.table` into `gwdetchar.io.html`
* Move some `gwsumm` JavaScript and CSS lines into `gwdetchar/share/`

A few related things are changed within `gwdetchar`:

* Use `gwdetchar.io.html.table` to construct tables, with an "Export to CSV" button for the list of packages
* PEP-8 compliance fixes for `gwdetchar.io.html`

This is related to #288.

cc @duncanmmacleod

### DIFF
--- a/gwdetchar/io/html.py
+++ b/gwdetchar/io/html.py
@@ -209,8 +209,8 @@ def new_bootstrap_page(base=os.path.curdir, lang='en', refresh=False,
         default: None
     """
     # get kwargs with sensible defaults
-    css = kwargs.get('css', CSS_FILES)
-    script = kwargs.get('script', JS_FILES)
+    css = CSS_FILES.extend(kwargs.get('css', []))
+    script = JS_FILES.extend(kwargs.get('script', []))
     # write CSS to static dir
     css, script = finalize_static_urls(
         os.path.join(os.path.curdir, 'static'),
--- a/gwdetchar/io/html.py
+++ b/gwdetchar/io/html.py
@@ -209,8 +209,8 @@ def new_bootstrap_page(base=os.path.curdir, lang='en', refresh=False,
         default: None
     """
     # get kwargs with sensible defaults
-    css = CSS_FILES.extend(kwargs.get('css', []))
-    script = JS_FILES.extend(kwargs.get('script', []))
+    css = CSS_FILES + kwargs.get('css', [])
+    script = JS_FILES + kwargs.get('script', [])
     # write CSS to static dir
     css, script = finalize_static_urls(
         os.path.join(os.path.curdir, 'static'),
--- a/gwdetchar/io/html.py
+++ b/gwdetchar/io/html.py
@@ -209,8 +209,8 @@ def new_bootstrap_page(base=os.path.curdir, lang='en', refresh=False,
         default: None
     """
     # get kwargs with sensible defaults
-    css = CSS_FILES + kwargs.get('css', [])
-    script = JS_FILES + kwargs.get('script', [])
+    css = kwargs.get('css', CSS_FILES)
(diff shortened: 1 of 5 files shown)


alurban:
@duncanmmacleod, I also plan to move gwsumm's navbar constructor into gwdetchar, but I'll do that in a separate PR because this one is already sizable (but well-contained).

gwdetchar/io/html.py
```diff
@@ -441,6 +441,71 @@ def write_arguments(content, start, end, flag=None, section='Parameters',
...
+    caption : `str`, optional
+        content for this table's `<caption>`
+
+    **class_
+        class attribute declarations for each tag used in the table,
+        any of `table`, `thead`, `tbody`, `tr`, `th`, `td`, `caption`
+
+    Returns
+    -------
+    table : `markup.page`
```
duncanmmacleod:
```suggestion
    table : `MarkupPy.markup.page`
```

gwdetchar/io/html.py
```diff
@@ -441,6 +441,71 @@ def write_arguments(content, start, end, flag=None, section='Parameters',
...
+    page.tbody(**kwargs['tbody'])
+    for row in data:
+        page.tr(**kwargs['tr'])
+        for td in row:
+            page.td(td, **kwargs['td'])
+        page.tr.close()
+    page.tbody.close()
+    page.table.close()
+    # add export button
+    if id is not None:
```
duncanmmacleod:
```suggestion
    if id:
```

(since you don't want to set an empty string as the `id` either

gwdetchar/io/html.py
```diff
@@ -209,8 +209,8 @@ def new_bootstrap_page(base=os.path.curdir, lang='en', refresh=False,
         default: None
     """
     # get kwargs with sensible defaults
-    css = kwargs.get('css', CSS_FILES)
-    script = kwargs.get('script', JS_FILES)
+    css = CSS_FILES + kwargs.get('css', [])
+    script = JS_FILES + kwargs.get('script', [])
```
duncanmmacleod: What's the motivation of this particular change? I think the original motivation was that if the user wants to give some CSS or JS files, they should give everything they need. The original code also avoids the issue of loading multiple versions of the same library if the user wishes to select a different version of, say, jquery.

alurban: @duncanmmacleod, I was thinking ahead to the specific use case of gwsumm, which will inherit some of these javascript and stylesheets from gwdetchar, but you're right, the proper way to do that is for _gwsumm_ to pass `css=CSS_FILES + ...` and similar for `script`.

alurban:
@duncanmmacleod, thanks for reviewing, I've addressed your comments above.

PR #298, opened 2019-04-07
Command-line module for gwdetchar.scattering

alurban:
This PR implements a command-line module for scattering investigations, which can be invoked via

```bash
$ python -m gwdetchar.scattering [OPTIONS] GPSTIME
```

Changes include:
* Move `gwdetchar.scattering` into a proper sub-module with its own `__init__.py`
* Add a module `gwdetchar/scattering/__main__.py`, which projects scattering fringe frequencies of all the default optic motion channels
* For those whose projected fringes are above a given threshold, produce plots comparing these to a Q-scan of _h(t)_
* Include unit tests for the above
* Add help text  pointing to the new module from `gwdetchar-scattering` output
* Typo fix to `gwdetchar.lasso.tests`

This fixes #296.

cc @duncanmmacleod, @jrsmith02, @andrew-lundgren

Issue #296, opened 2019-04-06
Enhance the gwdetchar scattering module


alurban:
There's been a lot of work on scattering recently by @andrew-lundgren and @jrsmith02, who have code (based on GWpy) that projects optic motion into _h(t)_ scattering fringes. [**Here**](https://github.com/jrsmith02/ligo-detchar/tree/master/scattering) is one prototype, though @andrew-lundgren has an [**interactive version**](https://git.ligo.org/andrew-lundgren/detchar-notebooks/blob/Local/Local/Scattering/ViewScatter) (requires `LIGO.ORG` credentials) that combines several data streams and plots a spectrogram.

We talked on Slack about migrating this into the `gwdetchar.scattering` module in such a way that you can run it on the command line with `__name__ == __main__`, like so:

```bash
$ python -m gwdetchar.scattering [OPTIONS] ARGUMENTS
```

This would be nice to have for gwdetchar-0.5.0, which will likely be the next release.

andrew-lundgren:
The main things that could improve the code would be:
 1. Better spectrogram, or use Omega scan
 2. Add in other auxiliary channels like angular motion of optics, or motions of ISIs and tables
 3. Maybe a better derivative filter, with a cutoff, if the current one is noisy? I've got a simple one.
 4. And of course a better command-line interface.

alurban:
Thanks @andrew-lundgren, I'm guessing points (2) and (3) also apply to the `gwdetchar-scattering` tool? It's easy to [**add channels**](https://github.com/gwdetchar/gwdetchar/blob/master/gwdetchar/scattering.py#L26-L62), and we can think about improving the derivative filter, which I think right now is just a linear difference between data points.

Issue #84, opened 2018-06-09
Add support for captions (with default text) in gwdetchar tools


alurban:
For clarity and ease of use, there's a push to add fancybox captions to figures that appear on the DetChar summary pages. It would be great if this idea extended to `gwdetchar` tools as well. I've already implemented this for Hveto:

https://github.com/hveto/hveto/pull/83

The same idea can be ported to `gwdetchar` in a relatively straightforward way. I'm happy to take this on over the summer as we prepare for pre-O3.

PR #301, opened 2019-04-11
More scattering upgrades

alurban:
This PR implements a few more scattering code upgrades:

* Include more longitudinal-motion OSEMs in the default channel list
* Do not display channels without evidence for scattering on the output page (but still plot them)
* Handle the use case where no evidence of scattering is found in non-empty observing segments
* Provide figure captions for scattering plots
* Include Omicron trigger central frequency and SNR in the scattering summary

An example run analyzing Livingston data from 2019-04-05 is available [**here**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/scattering/day/20190405/) (requires `LIGO.ORG` credentials).

Since these changes depend on the new `gwdetchar.scattering` module, I'm marking this a work-in-progress until #298 is merged.

This is related to #84 and #296.

cc @duncanmmacleod, @siddharth101, @jrsmith02, @andrew-lundgren


alurban:
@duncanmmacleod, now that #298 is merged, this PR is ready for review.

PR #310, opened 2019-04-14
Handle all context in get_command_line

alurban:
This PR re-organizes contextual paragraphs about command-line usage into the `get_command_line` utility. This makes it so that calling `gwdetchar.io.html.get_command_line()` returns HTML formatted like so:

- - -

This page was generated with the following command-line call:

```bash
$ gwdetchar-omega -i H1 1126259462
```

The install path used was `/path/to/install`.

- - -

cc @duncanmmacleod


duncanmmacleod:
Why not just use `sys.prefix`?

alurban:
@duncanmmacleod, the short answer is "because I wasn't aware of it," but I've availed myself and decided that using `sys.prefix` would simplify things. I've addressed this in the latest commit.

PR #312, opened 2019-04-15
Gracefully handle null use cases

alurban:
This PR fixes a bug I discovered in online processing, namely that `gwdetchar-overflow` and `gwdetchar-scattering` each fail when run with `--state-flag` over times when the given analysis flag is inactive. The issue is essentially an indexing error when attempting to range over an empty list.

So, the following changes are included:
* `gwdetchar-scattering`: if `state-flag` segments are empty, explicitly set the list of channels to an empty list
* `gwdetchar-overflow`: if overflow segments are all empty, handle that case separately, and provide a unit test

Example output is available for [**scattering**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/scattering/day/20190414/) and [**overflows**](https://ldas-jobs.ligo-la.caltech.edu/~aurban/overflows/day/20190414/) (both require `LIGO.ORG` credentials).

cc @duncanmmacleod

### DIFF
--- a/gwdetchar/utils.py
+++ b/gwdetchar/utils.py
@@ -121,10 +121,7 @@ def table_from_segments(flagdict, sngl_burst=False, snr=10., frequency=100.):
 
     for name, flag in flagdict.items():
         rows.extend(map(partial(row, channel=name), flag.active))
-    if rows:
-        table = EventTable(rows=rows, names=names)
-    else:
-        table = EventTable(names=names)
+    table = EventTable(rows=rows or None, names=names)
     if sngl_burst:  # add tablename for GWpy's ligolw writer
         table.meta["tablename"] = "sngl_burst"
     return table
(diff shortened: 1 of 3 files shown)


gwdetchar/utils.py
```diff
@@ -121,7 +121,10 @@ def row(seg, channel):
 
     for name, flag in flagdict.items():
         rows.extend(map(partial(row, channel=name), flag.active))
-    table = EventTable(rows=rows, names=names)
+    if rows:
+        table = EventTable(rows=rows, names=names)
+    else:
+        table = EventTable(names=names)
```
duncanmmacleod: You _could_ simplify this to

```python
    table = EventTable(rows=rows or None, names=names)
```

and remove the conditional, but the current way is probably more readable.

alurban: Thanks @duncanmmacleod, I like the way your suggestion takes up fewer lines, so I went ahead and implemented it.

PR #267, opened 2019-03-13
Automatically exclude missing channels

alurban:
This PR adds a new utility under `gwdetchar.io.datafind` called `remove_missing_channels`. Whenever a list of channels is requested to `get_data`, this tool is used to exclude those channels that are missing from the first or last frame file within the given cache, throwing a loud warning showing exactly which channels are being ignored.

Two things to note:
* If NDS is used instead, `remove_missing_channels` is not invoked, meaning the call will still fail if channels cannot be found over the server.
* If only a single channel is passed (not a list), but that channel cannot be found in requested local frame files, then the call will still fail.

**Upshot:** Since all tools use `get_data` to retrieve timeseries data, this utility will cut down on the number of jobs that fail because channels are requested that cannot be found for the given time range, _except_ when using NDS or reading one channel at a time. In the latter case, users can easily work around this by wrapping the function call in a `try-except` block.

This fixes #150.

cc @duncanmmacleod, @areeda, @tjma12, @macedo22, @uberpye, @jrsmith02

PR #324, opened 2019-05-12
Drop explicit support for python2

alurban:
This PR drops all support for python-2.7, now that we've been running stably in python-3.7:

* Remove imports from `__future__`
* Replace imports from `six` with their python3 equivalents
* Enforce system python versions >= 3.5

This is part of an extensive purge of python2 across our detchar software stack, see gwdetchar/hveto#126 and gwpy/gwsumm#267.

cc @duncanmmacleod


bin/gwdetchar-scattering
```diff
@@ -25,14 +25,11 @@ time please use the command-line module:
 `python -m gwdetchar.scattering --help`
 """
 
-from __future__ import division
-
-import os.path
 import re
 import random
+import os.path
```
duncanmmacleod: Why was this moved?

gwdetchar/utils.py
```diff
@@ -21,15 +21,10 @@
...
-    from io import StringIO
-    from html.parser import HTMLParser
-except:  # python 2.7
-    from cStringIO import StringIO
-    from HTMLParser import HTMLParser
-
 import numpy
+from io import StringIO
+from functools import partial
+from html.parser import HTMLParser
```
duncanmmacleod: Can you move these about the numpy import, since they are standard library modules?

duncanmmacleod: Can you move these above the `numpy` import, since they are standard library modules?

setup.py
```diff
@@ -31,6 +29,10 @@
 # local setup utilities
 from _setup_utils import CMDCLASS as cmdclass
 
+# enforce python version
+if sys.version < '3.5':
+    raise ImportError("Python versions older than 3.5 are not supported.")
+
```
duncanmmacleod: I don't think we need to include this.

versioneer.py
```diff
@@ -10,7 +10,7 @@
 * https://github.com/warner/python-versioneer
 * Brian Warner
 * License: Public Domain
-* Compatible With: python2.6, 2.7, 3.3, 3.4, 3.5, and pypy
+* Compatible With: python3.5+, conda, and pypy
```
duncanmmacleod: Please revert this change, since this is just a vendored copy of `versioneer`.

```suggestion
* Compatible With: python2.6, 2.7, 3.3, 3.4, 3.5, and pypy
```

bin/gwdetchar-scattering
alurban: @duncanmmacleod, mostly it's because I'm obsessive about the flow of text width in import statements -- I like to go from narrow to wide, for cleanliness of presentation

gwdetchar/utils.py
alurban: @duncanmmacleod, this is done in the last commit.

alurban:
@duncanmmacleod, thanks for your review, I've addressed your comments above

PR #330, opened 2019-05-29
Use large text in navbar branding

alurban:
This PR adjusts navbar brands (i.e., the text on the far-left, listed first before the nav menu) so that they are rendered with `<big>` HTML tags.

Example omega scan output is available [**here**](https://ldas-jobs.ligo.caltech.edu/~alexander.urban/wdq/Network_1187008882.0/) (requires `LIGO.ORG` credentials).

cc @duncanmmacleod, @areeda


alurban:
Thanks @areeda!

PR #342, opened 2019-06-17
Transmons for scattering segments.

siddharth101:
We have added the transmons channels particularly ASC-X_TR_B_NSUM_OUT_DQ and ASC-Y_TR_B_NSUM_OUT_DQ to find scattering, after having found strong evidence of these channel being good witness of the scattering noise. The scattering segments are generated using the band limited rms of the TimeSeries of transmons. These segments along with the segments generated from osems are being written in a hdf5 file, which can be downloaded from the scattering summary page.

### DIFF
--- a/bin/gwdetchar-scattering
+++ b/bin/gwdetchar-scattering
@@ -33,6 +33,8 @@ from io import StringIO
 
 import numpy
 
+from MarkupPy import markup
+
 from matplotlib import (use, rcParams)
 use('agg')  # noqa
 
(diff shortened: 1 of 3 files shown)


siddharth101:
@alurban I ran the script and the output is here [https://ldas-jobs.ligo-la.caltech.edu/~siddharth.soni/scatoutput/scattestoutput4/](url)
It looks good but for some reason, when I click on the Tranmons bar, it does not open.

PR #348, opened 2019-07-25
Add spectrum plots for band passes

macedo22:
This PR adds spectrum plots for when the band pass argument is used. It displays two plots:

- A plot that overlays the unfiltered and filtered spectra
- The same plot but zoomed in on the band pass limits, so as to visualize it up close

It also fixes the y axis range on scatter plots. When the units were extremely small, they were being plotted as a flat line. See an example of that behavior in any of the channel scatter plots here: https://ldas-jobs.ligo-wa.caltech.edu/~alexandra.macedo/detchar/LASSO/O3/48Hz-investigation/cleaned-strain/20190715/7am-9am/46-51Hz-minute/ 

**Example output**

With band pass: https://ldas-jobs.ligo-wa.caltech.edu/~alexandra.macedo/detchar/LASSO/example-output/spectra-plots/20190715/cleaned-strain/
Without band pass: https://ldas-jobs.ligo-wa.caltech.edu/~alexandra.macedo/detchar/LASSO/example-output/spectra-plots/20190715/bns-range/


gwdetchar/lasso/plot.py
```diff
@@ -72,3 +73,67 @@ def save_figure(fig, pngfile, **kwargs):
...
+    # Attributes shared between zoomed out and zoomed in plots
+    channelstub = re_delim.sub('_', str(channel_name)).replace('_', '-', 1)
+    gpsstub = '%d-%d' % (start, end-start)
+    x_label = 'Frequency [Hz]'
+    if ':GDS-CALIB_STRAIN' in channel_name:
+        y_label = 'GW amplitude spectral density [strain/$\sqrt{\mathrm{Hz}}]$'
+    else:
+        y_label = 'Primary channel units'
+    y_min = 1e-24
+    y_max = 2e-18
```
macedo22: These y axis limits are great for when strain is the primary channel, but would they need to be decided in a more general way for other possible primary channels?

gwdetchar/lasso/plot.py
```diff
@@ -72,3 +73,67 @@ def save_figure(fig, pngfile, **kwargs):
...
+    # Atttributes specific to one of the two spectrum plots
+    zoomed_out_x_min = 10
+    zoomed_out_x_max = 2000
+    zoomed_out_title = '%s band passed spectrum (zoomed out)' % channelstub
+    zoomed_out_spectrum_plot_name = ('%s-_BAND_PASSED_SPECTRUM_ZOOM_OUT-%s.png'
+                                     % (channelstub, gpsstub))
+
+    zoomed_in_x_min = flower - ((fupper - flower) * 0.1)
+    zoomed_in_x_max = fupper + ((fupper - flower) * 0.1)
+    zoomed_in_title = '%s band passed spectrum (zoomed in)' % channelstub
```
macedo22: Possible better titles for this and the zoomed out plot?

gwdetchar/lasso/plot.py
alurban: @macedo22, I see what you mean, how about something like this?

```python
    y_min = 0.1 * filtered_spectrum.min().value
    y_max = 5 * filtered_spectrum.max().value
```

I think we can live with this plot always showing up with a log scaled y-axis, but this change would make the axis limits more adaptive.

bin/gwdetchar-lasso-correlation
```diff
@@ -605,14 +625,27 @@ if args.band_pass:
 page.add(htmlio.parameter_table(content, start=start, end=end))
 page.add('<hr class="row-divider">')
 
+if args.band_pass:
+    page.h2('Primary shannel spectra', id_='spectra')
+    page.div(style_='display: flex;flex-wrap:nowrap;justify-content:space-around;')
```
alurban: @macedo22, inline style definitions like this are generally discouraged in favor of separate cascading style sheets (CSS), and same goes for the empty divs that follow. However, I think you can forego this problem altogether and use the standard plot scaffolding, something like:

```python
    page.h2('Primary shannel spectra', id_='spectra')
    page.div(class_='scaffold well')
    page.div(class_='row')
    page.div(class_='col-md-6')
    spectra_img1 = htmlio.FancyPlot(spectrum_plot_zoomed_out)
    page.add(htmlio.fancybox_img(spectra_img1))
    page.div.close()  # col-md-6
    page.div(class_='col-md-6')
    spectra_img2 = htmlio.FancyPlot(spectrum_plot_zoomed_in)
    page.add(htmlio.fancybox_img(spectra_img2))
    page.div.close()  # col-md-6
    page.div.close()  # row
    page.div.close()  # scaffold well
```

This structure takes full advantage of features native to twitter-bootstrap, so you have a scaffold well (the panel object with a drop-shadow) containing a row with two responsive, equal-size columns. (Note, "responsive" here means they will be formatted well on just about all screens regardless of size.)

gwdetchar/lasso/plot.py
alurban: @macedo22, the only thing I'd change is to put a hyphen in (`band-passed` rather than `band passed`), but only because I'm picky.

macedo22:
@alurban Thanks for the feedback! I believe I addressed each of your three comments, and the two example outputs I linked above have been rerun, so they reflect the new commits.

bin/gwdetchar-lasso-correlation
```diff
@@ -605,14 +625,29 @@ if args.band_pass:
...
 page.add('<hr class="row-divider">')
 
+if args.band_pass:
+    page.h2('Primary channel spectra', id_='spectra')
+    page.div(class_='scaffold well')
+    page.div(class_='row')
+    page.div(class_='col-md-6')
+    spectra_img1 = htmlio.FancyPlot(spectrum_plot_zoomed_out)
+    page.add(htmlio.fancybox_img(spectra_img1))
+    page.div.close()
```
alurban:
```suggestion
    page.div.close()  # col-md-6
```

bin/gwdetchar-lasso-correlation
```diff
@@ -605,14 +625,29 @@ if args.band_pass:
...
+    page.div(class_='col-md-6')
+    spectra_img2 = htmlio.FancyPlot(spectrum_plot_zoomed_in)
+    page.add(htmlio.fancybox_img(spectra_img2))
+    page.div.close()
```
alurban:
```suggestion
    page.div.close()  # col-md-6
```

bin/gwdetchar-lasso-correlation
```diff
@@ -605,14 +625,29 @@ if args.band_pass:
...
+    page.div.close()
```
alurban:
```suggestion
    page.div.close()  # row
```

bin/gwdetchar-lasso-correlation
```diff
@@ -605,14 +625,29 @@ if args.band_pass:
...
+    page.div.close()
```
alurban:
```suggestion
    page.div.close()  # scaffold well
```

macedo22:
Done! Apologies for omitting those details the first time. I agree that the practice of including them is extremely helpful.

PR #350, opened 2019-08-29
Datafind bug fix

dethodav:
This PR fixes a bug in how the `datafind.get_data()` module operates. If a frame type is provided, but no files are found by `gwdatafind`, the `source` variable is overridden and causes an index error. This can be avoided by ensuring that when no files are found, we reset `source = None`. 

An example of the error that currently occurs with the LIGO Conda ligo-py27 (and ligo-py37) environment is below:

```
RuntimeError: /home/derek.davis/.conda/envs/ligo-py27-clone/lib/python2.7/site-packages/gwdatafind/http.py:431: UserWarning: Missing segments: 
[1240215246.02 ... 1240215760.02)
warnings.warn(msg)
Traceback (most recent call last):
File "/home/derek.davis/.conda/envs/ligo-py27-clone/bin/gwdetchar-omega", line 221, in 
nproc=args.nproc, verbose='Reading block:'.rjust(30))
File "/home/derek.davis/.conda/envs/ligo-py27-clone/lib/python2.7/site-packages/gwdetchar/io/datafind.py", line 206, in get_data
channel = remove_missing_channels(channel, source)
File "/home/derek.davis/.conda/envs/ligo-py27-clone/lib/python2.7/site-packages/gwdetchar/io/datafind.py", line 114, in remove_missing_channels
available = set(io_gwf.iter_channel_names(gwfcache[0]))
IndexError: list index out of range
```

This can be reproduced by providing `gwdetchar-omega` a frame type for a gps time that does not exist.

Alternatively, this error could be fixed by changes to `gwdatafind` itself.

### DIFF
--- a/gwdetchar/io/datafind.py
+++ b/gwdetchar/io/datafind.py
@@ -198,6 +198,8 @@ def get_data(channel, start, end, frametype=None, source=None,
                 'Could not determine observatory from frametype')
         except (HTTPError, JSONDecodeError):  # frame files not found
             pass
+    if len(source) == 0: #if no frames found
+        source = None
     if isinstance(source, list) and isinstance(channel, (list, tuple)):
         channel = remove_missing_channels(channel, source)
     if source is not None:  # read from frame files
--- a/gwdetchar/io/datafind.py
+++ b/gwdetchar/io/datafind.py
@@ -198,7 +198,7 @@ def get_data(channel, start, end, frametype=None, source=None,
                 'Could not determine observatory from frametype')
         except (HTTPError, JSONDecodeError):  # frame files not found
             pass
-    if len(source) == 0: #if no frames found
+    if len(source) == 0: # reset if no frames found
         source = None
     if isinstance(source, list) and isinstance(channel, (list, tuple)):
         channel = remove_missing_channels(channel, source)
--- a/gwdetchar/io/datafind.py
+++ b/gwdetchar/io/datafind.py
@@ -198,11 +198,9 @@ def get_data(channel, start, end, frametype=None, source=None,
                 'Could not determine observatory from frametype')
         except (HTTPError, JSONDecodeError):  # frame files not found
             pass
-    if len(source) == 0: # reset if no frames found
-        source = None
-    if isinstance(source, list) and isinstance(channel, (list, tuple)):
+    if source and isinstance(source, list) and isinstance(channel, (list, tuple)):
         channel = remove_missing_channels(channel, source)
-    if source is not None:  # read from frame files
+    if source:  # read from frame files
         return series_class.read(
             source, channel, start=start, end=end, nproc=nproc,
             verbose=verbose, **kwargs)


dethodav:
To reproduce this error, I am able to run

```
conda activate ligo-py27
gwdetchar-omega 1187008882 -i L1 -o . -f L1-custom.ini 
```

Where `L1-custom` is

```
[GW]
name = Gravitational Wave Strain
q-range = 3.3166,150.0
frequency-range = 10.0,inf
resample = 4096
frametype = L1_HOFT_XXX
state-flag = L1:DMT-GRD_ISC_LOCK_NOMINAL:1
duration = 64
fftlength = 8
max-mismatch = 0.2
snr-threshold = 5
always-plot = True
plot-time-durations = 1,4,16
channels = L1:GDS-CALIB_STRAIN
```

However, removing the `frametype` line from the config allows the script to run to completion.

gwdetchar/io/datafind.py
```diff
@@ -198,6 +198,8 @@ def get_data(channel, start, end, frametype=None, source=None,
                 'Could not determine observatory from frametype')
         except (HTTPError, JSONDecodeError):  # frame files not found
             pass
+    if len(source) == 0: # reset if no frames found
+        source = None
     if isinstance(source, list) and isinstance(channel, (list, tuple)):
         channel = remove_missing_channels(channel, source)
     if source is not None:  # read from frame files
```
alurban: @dethodav, this can be done more simply by just changing line 203:

```suggestion
     if source:  # read from frame files
```

This logic would proceed to read from a frame cache only if `source` is neither `None`, `0`, nor an empty iterable, which should clear up the issue you're seeing.

gwdetchar/io/datafind.py
```diff
@@ -198,6 +198,8 @@ def get_data(channel, start, end, frametype=None, source=None,
                 'Could not determine observatory from frametype')
         except (HTTPError, JSONDecodeError):  # frame files not found
             pass
+    if len(source) == 0: # reset if no frames found
```
alurban: @dethodav, this attempt to fix the issue only introduces a different one, see the build test failures for an example. The issue is that this `source` variable is a `NoneType` object by default, so it has no length attribute.

However, I believe a simpler fix is possible through a one-line change on line 203.

gwdetchar/io/datafind.py
dethodav: I see the issue with my original commit, and this is definitely cleaner. However, there does still appear to be a needed change on line 201.

gwdetchar/io/datafind.py
```diff
@@ -198,6 +198,8 @@ def get_data(channel, start, end, frametype=None, source=None,
...
+        source = None
     if isinstance(source, list) and isinstance(channel, (list, tuple)):
```
dethodav: @alurban In order to prevent an empty list being passed, could we add an additional condition here?

```suggestion
    if source and isinstance(source, list) and isinstance(channel, (list, tuple)):
```

alurban: Aha, this is indeed a necessary change, good catch @dethodav. The call to `remove_missing_channels` needs to stay (it prevents job failures when expected channels are not found in frames), so we need this change to line 201 and line 203 should read:

```python
    elif source:
```

dethodav:
@alurban I've implemented the suggested changes to the PR. I did use your original suggestion for line 203, which now reads
```python
if source:
```
My understanding is that keeping it as `if source is none` would still allow an empty list be passed to the `read()` function, which would result in another failure.

PR #373, opened 2020-01-11
Add python3.8 support

duncanmmacleod:
This PR adds python3.8 support.


duncanmmacleod:
Ok, @alurban, this PR now reworks the entire Travis CI configuration to use `conda`, and implements OSX builds on Travis, as well as deprecating python3.5 and adding python3.8. This now reproduces the failure we saw in the gwdetchar-feedstock.

I think the python3.8/osx problem is that some tests use `os.chdir` to move into a temporary directory, then `shutil.rmtree` that directory without `os.chdir`ing back out again, so when `multiprocessing` attempts to spawn a child process, the current working directory doesn't actually exist.

alurban:
Thanks for digging into this @duncanmmacleod, I can confirm that is a thing that happens in several unit tests, do you want me to take a look at that? Maybe use a decorator to handle the staging of those directories?

duncanmmacleod:
>  do you want me to take a look at that? Maybe use a decorator to handle the staging of those directories?

I have an idea for a quick fix, give me a day or two to try that out.

duncanmmacleod:
Ok, @alurban, this is now ready. The really important changes are just

- adding `gwdetchar/conftest.py` to work around the directory issues in testing
- replace `python setup.py test` with `python -m pytest` in Travis, since pytest-runner doesn't seem to work with multiprocessing with spawn (default on macOS for python3.8)

Everything else is just ancillary stuff to support python3.8, or extras I put in whilst testing that can be preserved.

PR #381, opened 2020-08-24
Properly drop support for python < 3.6

### LINKED ISSUE #380: gwpy.testing.compat will no longer exist as of 2.0.0
See gwpy/gwpy#1244. If we aren't dependent on python-3.5 (i.e., if we only require python >= 3.6) then this is a non-issue as we can switch that module out for `unittest`.

alurban:
This PR properly drops support for python < 3.6:

* Drop package dependencies that rely on python < 3.5
* Drop imports from the deprecated `gwpy.testing.compat` module
* Simplify and refactor Conda builds
* Build-and-test only on the Linux platform with `language: minimal` (this anticipates refactoring `bin/` scripts to make GWDetChar into a `noarch: python` package)

This is related to #373 and fixes #380.


siddharth101:
@alurban I looked at the changed files and this all looks good to me.

duncanmmacleod:
I'm not the biggest fan of just copying the `parse-conda-requirements.py` script out of gwpy (without attribution). Is it safer to clone the gwpy repo on-the-fly to pick up the latest version - mainly so that changes don't have to be duplicated?

alurban:
@duncanmmacleod, fair enough, do you mean cloning it as part of the CI build? I can certainly do that

duncanmmacleod:
> @duncanmmacleod, fair enough, do you mean cloning it as part of the CI build? I can certainly do that

Yes, you should be able to do either

```
git clone --depth 1 https://github.com/gwpy/gwpy.git
```

to clone only the latest edition of the repo (no history), or just curl the 'raw' URL for that file: 

```
curl -LO https://raw.githubusercontent.com/gwpy/gwpy/master/ci/parse-conda-requirements.py
```

alurban:
Curling the raw file seems to have the least overhead while still guaranteeing to pick up the most recent changes, so I'll give that a shot. (I'll have to repeat this in hveto, which isn't a problem)

alurban:
@duncanmmacleod, while you're here, do you know why codecov doesn't seem to be interacting with this pull request? I'm having a hard time discerning whether I've changed a subtle setting somewhere or if it's a codecov issue

duncanmmacleod:
> @duncanmmacleod, while you're here, do you know why codecov doesn't seem to be interacting with this pull request? I'm having a hard time discerning whether I've changed a subtle setting somewhere or if it's a codecov issue

Check here: https://codecov.io/account/gh/gwdetchar

I get a 404 from that, meaning I'm not authorised to view that (feel free to add me as an administrator, if you think I can help).

alurban:
@duncanmmacleod, I get a 401 unauthorized from that link, in order to access settings I have to go in through my account portal. I've reset webhooks but am still perplexed, apparently a report was prepared but never sent because it never registered that the CI build was done...?

alurban:
Now we're cookin' with gas. I had to enable Codecov as a GitHub app, which is confusing, because coverage reports definitely worked before. At any rate, it works now, so I'm comfortable filing this under "whatever."

PR #385, opened 2020-10-18
Jlynam lasso changes branch

jlynam:
This pull request adds an additional method for removing outliers in `gwdetchar-lasso-correlation`:

- Adds arguments to both `find_outliers` and `remove_outliers` in `core.py` that allow for the specification of desired outlier removal method
- Adds appropriate unittests for the additions to `core.py`
- Adds new outlier removal parsing argument `gwdetchar-lasso-correlation`

### DIFF
--- a/bin/gwdetchar-lasso-correlation
+++ b/bin/gwdetchar-lasso-correlation
@@ -85,7 +85,7 @@ parser.add_argument('-p', '--primary-channel',
 parser.add_argument('-P', '--primary-frametype', default=None,
                     help='frametype for --primary-channel, default: guess '
                          'by channel name')
-parser.add_argument('-O', '--remove-outliers-s', type=float, default=None,
+parser.add_argument('-O', '--remove-outliers', type=float, default=None,
                     help='Std. dev. limit for removing outliers')
 parser.add_argument('-R', '--remove-outliers-pf', type=float, default=None,
                     help='Percent limit for removing outliers')
@@ -198,9 +198,9 @@ else:
                          verbose='Reading:'.rjust(30),
                          nproc=args.nproc).crop(start, end)
 
-if args.remove_outliers_s:
-    logger.debug("-- Removing outliers above %f sigma" % args.remove_outliers_s)
-    gwlasso.remove_outliers(primaryts, args.remove_outliers_s)
+if args.remove_outliers:
+    logger.debug("-- Removing outliers above %f sigma" % args.remove_outliers)
+    gwlasso.remove_outliers(primaryts, args.remove_outliers)
 elif args.remove_outliers_pf:
     logger.debug("-- Removing outliers in the bottom %f percent of data" % args.remove_outliers_pf)
     gwlasso.remove_outliers(primaryts, args.remove_outliers_pf, method='pf')
--- a/bin/gwdetchar-lasso-correlation
+++ b/bin/gwdetchar-lasso-correlation
@@ -85,8 +85,10 @@ parser.add_argument('-p', '--primary-channel',
 parser.add_argument('-P', '--primary-frametype', default=None,
                     help='frametype for --primary-channel, default: guess '
                          'by channel name')
-parser.add_argument('-O', '--remove-outliers', type=float, default=None,
+parser.add_argument('-O', '--remove-outliers-s', type=float, default=None,
                     help='Std. dev. limit for removing outliers')
+parser.add_argument('-R', '--remove-outliers-pf', type=float, default=None,
+                    help='Percent limit for removing outliers')
 parser.add_argument('-t', '--threshold', type=float, default=0.0001,
                     help='threshold for making a plot')
 
@@ -196,9 +198,14 @@ else:
                          verbose='Reading:'.rjust(30),
                          nproc=args.nproc).crop(start, end)
 
-if args.remove_outliers:
-    logger.debug("-- Removing outliers above %f sigma" % args.remove_outliers)
-    gwlasso.remove_outliers(primaryts, args.remove_outliers)
+if args.remove_outliers_s:
+    logger.debug("-- Removing outliers above %f sigma" % args.remove_outliers_s)
+    gwlasso.remove_outliers(primaryts, args.remove_outliers_s)
+elif args.remove_outliers_pf:
+    logger.debug("-- Removing outliers in the bottom %f percent of data" % args.remove_outliers_pf)
+    gwlasso.remove_outliers(primaryts, args.remove_outliers_pf, method='pf')
+    start = int(primaryts.span()[0])
+    end = int(primaryts.span()[1])
 
 primary_mean = numpy.mean(primaryts.value)
 primary_std = numpy.std(primaryts.value)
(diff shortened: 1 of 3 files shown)


alurban:
@jlynam, thanks for posting this, and welcome to the team! Just checking in to acknowledge that I saw this, I'll try to post a review tomorrow, if you don't hear back by Tuesday please feel free to ping me.

bin/gwdetchar-lasso-correlation
```diff
@@ -85,8 +85,10 @@ parser.add_argument('-p', '--primary-channel',
 parser.add_argument('-P', '--primary-frametype', default=None,
                     help='frametype for --primary-channel, default: guess '
                          'by channel name')
-parser.add_argument('-O', '--remove-outliers', type=float, default=None,
+parser.add_argument('-O', '--remove-outliers-s', type=float, default=None,
```
alurban, line 88: I suggest we keep the original name of this argument, so that end users don't have to change any script invocations

gwdetchar/lasso/core.py
```diff
@@ -33,61 +33,91 @@
 
 # -- utilities ----------------------------------------------------------------
 
-def find_outliers(ts, N=5):
+def find_outliers(ts, L=5, method='s'):
```
alurban, line 36: I suggest we keep the `N` kwarg nomenclature; downstream users will see their code break otherwise

gwdetchar/lasso/core.py
```diff
@@ -33,61 +33,91 @@
...
 
     Parameters
     ----------
     ts : `~gwpy.timeseries.TimeSeries`
         data to find outliers within
 
-    N : `float`, optional
-        number of standard deviations to consider an outlier, default: 5
+    L : `float`, optional
+        if S is given as 's': number of standard deviations to consider an outlier, default: 5
```
alurban, line 45: Based on the kwarg names in the function definition, this should probably read something like:

```suggestion
        if ``method='s'``: number of standard deviations to consider an outlier, default: 5
```

etc.

Note: These comments apply to all changed function definitions and docstrings

gwdetchar/lasso/tests/test_core.py
```diff
@@ -51,18 +59,35 @@
 
 # -- unit tests ---------------------------------------------------------------
 
-def test_find_outliers():
```
alurban, line 54: I suggest keeping this unit test name the same, though it is a good idea to add the commend line below. Line 64 of this diff, which reads `# find expected outliers`, can be deleted in that case

gwdetchar/lasso/tests/test_core.py
```diff
@@ -51,18 +59,35 @@
...
+def test_find_outliers_s():
+    # test for standard deviation outlier finding
     # find expected outliers
     outliers = core.find_outliers(OUTLIER_TS)
     assert isinstance(outliers, numpy.ndarray)
     nptest.assert_array_equal(outliers, numpy.array([IND]))
 
 
-def test_remove_outliers():
```
alurban, line 61: Same comments as in `test_find_outliers`

alurban (changes requested):
Thanks for this contribution @jlynam! The lasso codebase will definitely be improved by having this extra way to find and remove outliers.

The overall code changes you have look good to me, however, I have a few very minor suggestions which mostly revolve around making sure existing command-line flags, function names, and keyword arguments (kwargs) stay the same to avoid breaking any downstream code.

Apart from that, there are some PEP-008 convention complaints from the `flake8` linter which should be addressed. It looks like these are mostly lines longer than 79 characters, which are annoying but very easy to fix. I recommend installing the `flake8` package and using it to check your branch. From the top level of the repo, you can do this on the command-line via:

```bash
$ python -m flake8 gwdetchar/
```

bin/gwdetchar-lasso-correlation
```diff
@@ -199,6 +201,11 @@ else:
 if args.remove_outliers:
     logger.debug("-- Removing outliers above %f sigma" % args.remove_outliers)
     gwlasso.remove_outliers(primaryts, args.remove_outliers)
+elif args.remove_outliers_pf:
+    logger.debug("-- Removing outliers in the bottom %f percent of data" % args.remove_outliers_pf)
```
alurban, line 205: The `flake8` linter is complaining that this line is too long, can you break it up across two lines and use string formatting like so?
```suggestion
    logger.debug("-- Removing outliers in the bottom {} percent "
                 "of data".format(args.remove_outliers_pf))
```

gwdetchar/lasso/core.py
```diff
@@ -42,18 +42,32 @@ def find_outliers(ts, N=5):
         data to find outliers within
 
     N : `float`, optional
-        number of standard deviations to consider an outlier, default: 5
+        if `method='s'`: number of standard deviations to consider an outlier
+        if `method='pf'`: percentile range limit to consider an outlier
+        default value for both methods:
+    S : `String`, optional
```
alurban, line 48: This part of the docstring should reflect the name of the keyword argument in the function spec, something like:

```python
    method : `str`, optional
        outlier identification method to use, must be either `'s'` (standard
        deviation method) or `'pf'` (percentile range method),
        default: `'s'`
```

gwdetchar/lasso/core.py
```diff
@@ -42,18 +42,32 @@ def find_outliers(ts, N=5):
         data to find outliers within
 
     N : `float`, optional
-        number of standard deviations to consider an outlier, default: 5
+        if `method='s'`: number of standard deviations to consider an outlier
+        if `method='pf'`: percentile range limit to consider an outlier
+        default value for both methods:
```
alurban, line 47: This seems to be missing a default value:

```suggestion
        default for both methods: 5
```

gwdetchar/lasso/core.py
```diff
@@ -62,31 +76,51 @@ def remove_outliers(ts, N=5):
         data to find outliers within
 
     N : `float`, optional
-        number of standard deviations to consider an outlier, default: 5
+        if `method='s'`: number of standard deviations to consider an outlier
+        if `method='pf'`: percentile range limit to consider an outlier
+        default value for both methods: 5
+    S : `String`, optional
```
alurban, line 82: Same comments as the docstring for `find_outliers`

alurban (changes requested):
Thanks again @jlynam! I took another pass through, it looks like you caught most of the PEP-008 complaints, I just have a couple of other (very minor) comments.

gwdetchar/lasso/core.py
```diff
@@ -62,31 +77,52 @@ def remove_outliers(ts, N=5):
         data to find outliers within
 
     N : `float`, optional
-        number of standard deviations to consider an outlier, default: 5
+        if `method='s'`: number of standard deviations to consider an outlier
+        if `method='pf'`: percentile range limit to consider an outlier
+        default for both methods: 5
+    method : `String`, optional
```
alurban, line 83: Same comment as in `find_outliers`

alurban (changes requested):

alurban (approved):
LGTM, thanks @jlynam

PR #389, opened 2020-12-13
Making sure the Coefficients and Channels match the right columns

jlynam:
The "Top Channels" cluster csv files are formatted incorrectly. This merge swaps the two column names, 'Channel' and 'Pearson Coefficient', so that they match the appropriate columns.


alurban:
LGTM, thanks @jlynam, this will help avoid confusion in the future

PR #390, opened 2021-01-13
Address race condition in logger

duncanmmacleod:
Sometimes (I can't nail down _when_ this will happen) `coloredlogs.install()` will set the logging level on the parent logger _only_ and not the actual logger we gave it, so this PR adds a safety net to force the new logger to have the correct level.

cf https://git.ligo.org/computing/conda/-/jobs/1146040#L1527


alurban (approved):
LGTM, thanks @duncanmmacleod

PR #396, opened 2021-04-26
Pin parse-conda-requirements.py to a specific version

duncanmmacleod:
This PR pins the download reference for the `parse-conda-requirements.py` script, since the upstream gwpy is changing (see gwpy/gwpy#1320).

PR #397, opened 2021-06-10
Mock out package_list() function in all tests

duncanmmacleod:
This PR adds an `autouse` fixture to all tests to mock out the `gwdetchar.io.html.package_list` function, which removes unnecessary (and slow) `subprocess` calls out to `conda list` or `pip list installed`. There is a hook to allow tests of the `package_list` function itself to not receive the mock.

On my local machine on python3.6 on a single CPU installed as part of a very large environment (`igwn-py36-testing`), this reduces the overall runtime of the test suite by 40%.

### DIFF
--- a/gwdetchar/conftest.py
+++ b/gwdetchar/conftest.py
@@ -20,8 +20,45 @@
 """
 
 import os
+from unittest import mock
 
 import pytest
+
+
+def fake_package_list():
+    """Fake return for `gwdetchar.io.html.package_list`
+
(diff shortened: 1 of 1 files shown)


duncanmmacleod:
The CI is dependent upon #396.

eagoetz:
@duncanmmacleod What is the status of this PR; is it ready to be reviewed? The CI checks would need to be retriggered. Thanks

duncanmmacleod:
@eagoetz, this is ready.

PR #395, opened 2021-04-12
Fix AttributeError in gwdetchar.lasso.plot.configure_mpl_tex

duncanmmacleod:
This PR fixes #394 by modifying the imports of `matplotlib.texmanager` to prevent an `AttributeError`.

PR #399, opened 2021-06-10
Migrate CI to GitHub actions

duncanmmacleod:
This PR removes the Travis CI configuration and replaces it with one for Github Actions.


duncanmmacleod:
This needs #395 to be fixed first.

duncanmmacleod:
And in case anyone's wondering, this is deliberately posted from a branch of gwdetchar/gwdetchar, and not duncanmmacleod/gwdetchar because github actions doesn't run from forks if the target branch doesn't already have a github actions configuration on it, so this is the only way to test the CI config before merging.

eagoetz (approved):
As far as I can tell, having no experience with GitHub actions, this LGTM. I'm not sure the main motivation for switching the build-test-doc CI to GitHub actions, but I'm guessing @duncanmmacleod has already understood what the main reasons are. :wink: Again, LGTM.

PR #400, opened 2021-06-18
Remove unused lalsuite requirement

duncanmmacleod:
This PR removes the (apparently) unused `lalsuite` `install_requires` entry. I think the last need for this was removed in fa60c53bb939b92d16f2104bac576f597c3f775b.


eagoetz (approved):
LGTU

PR #401, opened 2021-06-25
Implemented choosing file to load primary channel from

wutmeownow:
Added get_primary_ts function in __main__.py of lasso to either fetch primary channel normally or load it from a given file path. This file path is given in the -pf argument for the parser. Also added a test_main.py file for the lasso __main__, although it only currently tests my get_primary_ts function. It's been tested with and without bandpass.

### DIFF
--- a/gwdetchar/lasso/__main__.py
+++ b/gwdetchar/lasso/__main__.py
@@ -21,6 +21,9 @@ import os
 import re
 import sys
 
+# added import
+from gwpy.timeseries import TimeSeries
+
 import numpy
 
 from MarkupPy import markup
@@ -320,7 +323,7 @@ def create_parser():
         '-p',
         '--primary-channel',
         default='{ifo}:DMT-SNSH_EFFECTIVE_RANGE_MPC.mean',
-        help='name of primary channel to use or filepath to primary timeseries .gwf file',
+        help='name of primary channel to use or filepath to primary timeseries .gwf file with channel name separated by space',
     )
     parser.add_argument(
         '-P',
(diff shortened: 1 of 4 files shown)


gwdetchar/lasso/__main__.py
```diff
@@ -425,9 +445,14 @@ def main(args=None):
     global nonzerocoef, nonzerodata, p1, primary, primary_mean, primary_std
     global primaryts, range_is_primary, re_delim, start, target, times
     global threshold, trend_type, xlim
-
+
```
eagoetz, line 448: Remove whitespace

gwdetchar/lasso/__main__.py
```diff
@@ -344,7 +364,7 @@ def create_parser():
         '--remove-outliers-pf',
         type=float,
         default=None,
-        help='Percent limit for removing outliers',
+        help='Percent limit for removing outliers between 0 and 1',
```
eagoetz, line 367: Instead of "Percent" suggest using "Fractional"

gwdetchar/lasso/__main__.py
```diff
@@ -267,6 +269,17 @@ def _process_channel(input_):
     return (chan, lassocoef, plot4, plot5, plot6, ts)
 
 
+def get_primary_ts(channel, start, end, filepath=None, frametype=None, cache=None, nproc=1):
```
eagoetz, line 272: I suggest adding some documentation and/or other comment to explain what this function is doing, even if it is straightforward

eagoetz:
@wutmeownow are there other comparison tests you can provide of a before-and-after nature, just to demonstrate that the modifications don't interfere with normal lasso operations?

gwdetchar/lasso/__main__.py
wutmeownow: Done!

gwdetchar/lasso/__main__.py
wutmeownow: Also done!

wutmeownow:
> @wutmeownow are there other comparison tests you can provide of a before-and-after nature, just to demonstrate that the modifications don't interfere with normal lasso operations?

I will go ahead and get up two (or more) lasso index pages for before/after.

eagoetz:
@wutmeownow Thanks for the updates! We looked at the code and it looks fine, but before we mark it approved and merge this, we wanted to clarify on the output result table: some of the correlation coefficient numerical values are slightly different. Is this to be expected? We were thinking that the outputs of at least the [original](https://ldas-jobs.ligo-la.caltech.edu/~michaeljohn.lowry/lasso_comparison/original_gwdetchar/) and [new version](https://ldas-jobs.ligo-la.caltech.edu/~michaeljohn.lowry/lasso_comparison/normal_query/) queries should be identical, but they are not. Thanks for helping to clarify this for us.

wutmeownow:
> @wutmeownow Thanks for the updates! We looked at the code and it looks fine, but before we mark it approved and merge this, we wanted to clarify on the output result table: some of the correlation coefficient numerical values are slightly different. Is this to be expected? We were thinking that the outputs of at least the [original](https://ldas-jobs.ligo-la.caltech.edu/~michaeljohn.lowry/lasso_comparison/original_gwdetchar/) and [new version](https://ldas-jobs.ligo-la.caltech.edu/~michaeljohn.lowry/lasso_comparison/normal_query/) queries should be identical, but they are not. Thanks for helping to clarify this for us.
> 
> The [new, custom primary channel version](https://ldas-jobs.ligo-la.caltech.edu/~michaeljohn.lowry/lasso_comparison/from_file/) also has a different output in the results table.

I believe that is to be expected because running it with the original multiple times also seems to come up with different coefficients as well, although they are usually around the same value. This is actually something Marissa was talked to me about and how she might like to investigate why this happens, but for now I would just expect it to be some random deviation. Thanks for pointing that out!

eagoetz:
I see. So if I understand properly, if you run the code with the exact same input data (all the same argument values, etc.) then you find numerical difference each time the program runs, correct? That seems quite unusual.

wutmeownow:
Exactly. It is weird but since the results are almost identical and the channels are in the same relative order, it isn't a huge issue. Marissa probably knows more about it if you are interested, but rest assured it's not caused by my changes.

PR #405, opened 2021-09-24
swap to gwpy-based datafind

dethodav:
This pull request swaps out the datafind method in gwdetchar from `gwdatafind.find_urls()` to `gwpy.io.datafind.find_urls()`. The relevant gwpy function can be seen here: 

https://github.com/gwpy/gwpy/blob/223e6e5fc9223c5b08975c1c023286546fe5094f/gwpy/io/datafind.py#L698-L709

As linked above, in the LIGO case, the function `gwdatafind.find_urls()` is still used. However, this function also has support for finding Virgo FFL urls. Hence, this PR is necessary for Virgo support by gwdetchar tools and should still have the same result in the LIGO case.

### DIFF
--- a/gwdetchar/io/datafind.py
+++ b/gwdetchar/io/datafind.py
@@ -24,8 +24,6 @@ import warnings
 from urllib.error import HTTPError
 from json.decoder import JSONDecodeError
 
-import gwdatafind
-
 from ..const import DEFAULT_SEGMENT_SERVER
 
 from gwpy.io import gwf as io_gwf
--- a/gwdetchar/io/datafind.py
+++ b/gwdetchar/io/datafind.py
@@ -29,6 +29,7 @@ import gwdatafind
 from ..const import DEFAULT_SEGMENT_SERVER
 
 from gwpy.io import gwf as io_gwf
+from gwpy.io import datafind as io_datafind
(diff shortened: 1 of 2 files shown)


gwdetchar/io/datafind.py
```diff
@@ -192,7 +193,7 @@ def get_data(channel, start, end, frametype=None, source=None,
         try:  # locate frame files
             ifo = re.search('[A-Z]1', frametype).group(0)
             obs = ifo[0]
-            source = gwdatafind.find_urls(obs, frametype, start, end)
+            source = io_datafind.find_urls(obs, frametype, start, end)
```
eagoetz, line 196: Now that this is swapped, is `gwdatafind` used anywhere else? Should this be imported anymore?

eagoetz:
@dethodav Aside from my minor comment about the `gwdatafind` import, it will be necessary to rebase this PR against the latest upstream version in the repo to confirm the CI checks all still pass

gwdetchar/io/datafind.py
dethodav: You're right that `gwdatafind` doesn't need to be imported anymore. I've gone ahead and removed the import statement

PR #406, opened 2021-09-24
Swap out broken link

dethodav:
This pull request swaps out a broken link that is hard-coded in the `gwdetchar` codebase. 

In the long term, if this feature is intended to be kept, this link should be a configurable option. In the short term, I suggest swapping to a link that we believe will be stable (e.g. my suggestion in the PR for astro pic of the day). If this is also a concern, then I suggest we simply remove the feature (and the link) until this code can be re-added.


eagoetz:
@dethodav Thanks for spotting this issue and providing an updated link. I agree we should figure out a better solution, but perhaps it's low priority work for the time being. Can you open an issue about that so that we don't forget about it?

Also, just for sanity's sake, can you rebase against the upstream so that the CI tests should ideally pass and then we can merge this PR? Thanks!

eagoetz (approved):
LGTM

PR #416, opened 2022-08-23
Fix CI build failures with latest pygments and rstcheck

### LINKED ISSUE #409: test_about_this_page fails with latest pygments
In an environment that includes the latest version of pygments (2.11.x), the `test_about_this_page` test function fails because the pygments output has changed:

```console
======================================== test session starts =========================================
platform linux -- Python 3.10.0, pytest-6.2.5, py-1.11.0, pluggy-1.0.0
rootdir: /home/duncan/git/gwdetchar, configfile: setup.cfg
plugins: xdist-2.4.0, requests-mock-1.9.3, forked-1.3.0, cov-3.0.0
collected 27 items

gwdetchar/io/tests/test_html.py .......F...................                                    [100%]

============================================== FAILURES ==============================================
________________________________________ test_about_this_page ________________________________________

package_list = <MagicMock name='package_list' id='140482595650720'>
tmpdir = local('/tmp/pytest-of-duncan/pytest-690/test_about_this_page0')

    @mock.patch(
        "gwdetchar.io.html.package_list",
        return_value=[
            {"name": "gwpy", "version": "1.0.0"},
            {"name": "gwdetchar", "version": "1.2.3"},
        ],
    )
    def test_about_this_page(package_list, tmpdir):
        outdir = str(tmpdir)
        config_file = os.path.join(outdir, 'test.ini')
        with open(config_file, 'w') as fobj:
            fobj.write(TEST_CONFIGURATION)
        testargs = ['/opt/bin/gwdetchar-scattering',
                    '-i', 'X1']
        with mock.patch.object(sys, 'argv', testargs):
            # test with a single config file
            about = html.about_this_page(config_file)
>           assert parse_html(about) == parse_html(ABOUT)
E           AssertionError: assert 'Start tag: d...nd tag: div\n' == 'Start tag: d...nd tag: div\n'
E             Skipping 1107 identical leading characters in diff, use -v to show
E               tag: span
E             + Start tag: span
E             + attr: ('style', 'color: #bbbbbb')
E             + End tag: span
E               Data:
E               ...
E
E             ...Full output truncated (87 lines hidden), use '-vv' to show

gwdetchar/io/tests/test_html.py:332: AssertionError
```

eagoetz:
This PR does several things
- Drops testing of python3.6
- Adds testing of python3.10
- Handles CI build failures from versions of `pygments-2.11` and higher
- Fixes `rstcheck` tests since version 6 has some API changes
- Fixes some flake8 failures in Lasso

Closes #409

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -45,8 +45,8 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install \
-            "rstcheck" \
+        python -m pip install "sphinx"
+        python -m pip install "rstcheck"
         ;
     - name: Lint with rstcheck
       run: rstcheck . --recursive
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -47,6 +47,5 @@ jobs:
         python -m pip install --upgrade pip
         python -m pip install "sphinx"
         python -m pip install "rstcheck"
-        ;
     - name: Lint with rstcheck
       run: rstcheck . --recursive
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -46,6 +46,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         python -m pip install "sphinx"
-        python -m pip install "rstcheck"
+        python -m pip install "rstcheck>=6.0"
     - name: Lint with rstcheck
       run: rstcheck . --recursive
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -49,4 +49,4 @@ jobs:
             "rstcheck" \
         ;
     - name: Lint with rstcheck
-      run: python -m rstcheck . --recursive
+      run: rstcheck . --recursive
--- a/gwdetchar/io/tests/test_html.py
+++ b/gwdetchar/io/tests/test_html.py
@@ -27,9 +27,9 @@ import sys
 
 from getpass import getuser
 from MarkupPy import markup
+from pygments import __version__ as pygments_version
 from pytz import reference
 from unittest import mock
-from pygments import __version__ as pygments_version
 
 from gwpy.segments import (Segment, DataQualityFlag)
 
--- a/gwdetchar/io/tests/test_html.py
+++ b/gwdetchar/io/tests/test_html.py
@@ -29,6 +29,7 @@ from getpass import getuser
 from MarkupPy import markup
 from pytz import reference
 from unittest import mock
+from pygments import __version__ as pygments_version
 
 from gwpy.segments import (Segment, DataQualityFlag)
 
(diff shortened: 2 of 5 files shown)


gwdetchar/io/tests/test_html.py
```diff
@@ -90,6 +90,24 @@
...
+<h2>Configuration files</h2>
+<p>The following INI-format configuration file(s) were passed on the comand-line and are reproduced here in full:</p>
+<div class="highlight" style="background: #f8f8f8"><pre style="line-height: 125%;"><span></span><span style="color: #008000; font-weight: bold">[section]</span><span style="color: #bbbbbb"></span>
+<span style="color: #687822">key</span><span style="color: #bbbbbb"> </span><span style="color: #666666">=</span><span style="color: #bbbbbb"> </span><span style="color: #BA2121">value</span><span style="color: #bbbbbb"></span>
+</pre></div>
+
+<h2 class="mt-4">Environment</h2><table class="table table-sm table-hover table-responsive" id="package-table"><caption>Table of packages installed in the production environment</caption><thead><tr><th scope="col">Name</th><th scope="col">Version</th></tr></thead><tbody><tr><td>gwdetchar</td><td>1.2.3</td></tr><tr><td>gwpy</td><td>1.0.0</td></tr></tbody></table><button class="btn btn-outline-secondary btn-table mt-2" data-table-id="package-table" data-filename="package-table.csv">Export to CSV</button>
+</div>
+</div>""".format(sys.prefix)  # noqa: E501
```
duncanmmacleod, line 109: If the reason for `ABOUT2` is because of changes in pygments, I would parametrise this based on the discovered version of pygments.

Something like:

```python
from pygments import __version__ as pygments_version

if pygments_version >= "2.11.0":
    pygments_output = (
        '<span style="color: #687822">key</span>'
        '<span style="color: #bbbbbb"> </span>'
        '<span style="color: #666666">=</span>'
        '<span style="color: #bbbbbb"> </span>'
        '<span style="color: #BA2121">value</span>'
        '<span style="color: #bbbbbb"></span>'
    )
else:
    pygments_output = (
        '<span style="color: #7D9029">key</span> '
        '<span style="color: #666666">=</span> '
        '<span style="color: #BA2121">value</span>'
    )

ABOUT = f"""<div class="row">
<div class="col-md-12">
<h2>On the command-line</h2>
<p>This page was generated with the following command-line call:</p>
<div class="highlight" style="background: #f8f8f8"><pre style="line-height: 125%;"><span></span>$ gwdetchar-scattering -i X1
</pre></div>
<p>The install path used was <code>{sys.prefix}</code>.</p>
<h2>Configuration files</h2>
<p>The following INI-format configuration file(s) were passed on the comand-line and are reproduced here in full:</p>
<div class="highlight" style="background: #f8f8f8"><pre style="line-height: 125%;"><span></span><span style="color: #008000; font-weight: bold">[section]</span>
{pygments_output}
</pre></div>
<h2 class="mt-4">Environment</h2><table class="table table-sm table-hover table-responsive" id="package-table"><caption>Table of packages installed in the production environment</caption><thead><tr><th scope="col">Name</th><th scope="col">Version</th></tr></thead><tbody><tr><td>gwdetchar</td><td>1.2.3</td></tr><tr><td>gwpy</td><td>1.0.0</td></tr></tbody></table><button class="btn btn-outline-secondary btn-table mt-2" data-table-id="package-table" data-filename="package-table.csv">Export to CSV</button>
</div>
</div>"""  # noqa: E501
```

eagoetz: Fixed

eagoetz:
@duncanmmacleod Should be good to go now

robertbruntz:
@eagoetz @duncanmmacleod Do we want to take Python 3.7 out of testing, as well?

eagoetz:
> @eagoetz @duncanmmacleod Do we want to take Python 3.7 out of testing, as well?

I don't have a preference. I left 3.7 in because can see in the git history that there was an obvious effort to remove python3.6 from the supported python versions. This PR just finishes that effort

robertbruntz:
Since 3.7 is [no longer supported in IGWN Conda environments](https://computing.docs.ligo.org/conda/environments/#python-versions), and we already have a PR open that is taking out 3.6, this seems like a good time to remove it.  @duncanmmacleod ?

duncanmmacleod:
@robertbruntz, @eagoetz, IMO you should take one of two roads:

-   Support all Python versions you can until doing so presents an unreasonable maintenance burden. Currently there is no good reason _not_ to support Python 3.7, since the code and its dependencies currently support it (the CI passes). But, as soon as someone wants to add a feature that doesn't work in Python <3.8 (e.g. assignment operators `:=`), or upgrades a dependency past the point that _it_ is supported on Python <3.8 (e.g Astropy >=5.0.0), you should stop supporting 3.7. This is the conservative approach.

OR

-   Follow (something like) [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) and actively only support Python versions that are supported upstream by your requirements (Python, Numpy, Astropy etc). This is more aggressive, but may present a lower maintenance burden as a result of having to support a smaller set of platforms at any given time. The IGWN Conda Distribution follows NEP-29 for this reason.

eagoetz:
@duncanmmacleod @robertbruntz I think we'll leave the decision about python3.7 support until another time. We can discuss next computing call, but in the meantime I think this PR is ready to be approved. Any further comments? Thanks!

robertbruntz:
There are a number of very specific changes that I don't understand or can only infer the reason for.  Normally I'd want to discuss those with Evan, but in the interest of time, @duncanmmacleod - can you review the changes and approve them (or not), so this can move forward so [the other issue](https://github.com/gwdetchar/gwdetchar/issues/409) can move forward so [the original PR](https://github.com/gwdetchar/gwdetchar/pull/414) can move forward?

.github/workflows/lint.yml
```diff
@@ -45,8 +45,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        python -m pip install \
-            "rstcheck" \
-        ;
+        python -m pip install "sphinx"
+        python -m pip install "rstcheck>=6.0"
```
duncanmmacleod, line 49: Why not

```suggestion
        python -m pip install \
            "rstcheck>=6.0" \
            "sphinx" \
        ;   
```

gwdetchar/io/tests/test_html.py
```diff
@@ -29,6 +29,7 @@
 from MarkupPy import markup
 from pytz import reference
 from unittest import mock
+from pygments import __version__ as pygments_version
```
duncanmmacleod, line 32: Since someone went to the trouble of importing things alphabetically with some grouping, I recommend preserving that for readability (ie move this line up two lines).

.github/workflows/lint.yml
eagoetz: Somehow this didn't work for me when I tried it. It threw up an error. Or maybe I did something wrong...

gwdetchar/io/tests/test_html.py
eagoetz: Done

PR #419, opened 2023-02-03
Batch omega scans need to set request_disk in condor submit files

areeda:
Very small update to allow condor batch processing.The merge commit message has sll of my merges, they can ALL be removed from the merge.


eagoetz:
@areeda I'm guessing this is an important fix. Could you have another look at this to get the CI build to pass?

eagoetz (approved):
LGTM

PR #420, opened 2023-02-25
Allow kwargs to pass through to find_urls() method

eagoetz:
This permits a user to specify options like on_gaps='warn', for example, to find_urls()


areeda (approved):
LGTM

Issue #403, opened 2021-09-22
Optionally provide Lasso a random seed value


eagoetz:
There appears to be some randomness in Lasso runs, probably because there is some random number process happening internally. For testing and debugging purposes, it would be highly beneficial for Lasso to be reproducible. This means a fixed seed value should optionally be provided so that running on the same data twice would yield the same results. Otherwise it is hard to figure out if code changes actually impacted (intended or not) Lasso calculations.

eagoetz:
Fixed

PR #422, opened 2023-03-29
Fixed random results in Lasso

### LINKED ISSUE #367: gwdetchar-lasso-correlation on Python 3 returning warnings on numerical issues
I noticed that gwdetchar-lasso-correlations **returns different answers** for the same run parameters on 
Python 2: https://ldas-jobs.ligo-wa.caltech.edu/~jrsmith/detchar/O3b/lasso/
vs Python 3: https://ldas-jobs.ligo-wa.caltech.edu/~jrsmith/detchar/O3b/lasso-Nov-6-2019-2/

Additionally, with Python 3 it returns many more warnings. It seems like there are about a billion of these two different warnings for Python 3:  

  warnings.warn("Numerical issues were encountered "
/home/detchar/.conda/envs/ligo-summary-3.7/lib/python3.7/site-packages/sklearn/preprocessing/data.py:172: UserWarning: Numerical issues were encountered when centering the data and might not be solved. Dataset may contain too large values. You may need to prescale your features.

  warnings.warn("Numerical issues were encountered "
/home/detchar/.conda/envs/ligo-summary-3.7/lib/python3.7/site-packages/sklearn/preprocessing/data.py:189: UserWarning: Numerical issues were encountered when scaling the data and might not be solved. The standard deviation of the data is probably very close to 0. 

There was only one such warning for the Python 2 warning. Perhaps a clue for the channel differences seen.

lukecotnoir:
When running lasso multiple times with the same data set, the results were slightly different because the method used to fetch auxiliary channel data, get_data(), returns a time series dictionary with channels in random order. Because the default mode of the lasso regression is cyclic, this would produce random results. To fix this, the dictionary can be re-ordered to match the order of the auxiliary channel list to keep results repeatable. 

This PR also fixes issues with plotting when using newer versions of gwpy, matplotlib, python, and fixes warnings

Closes #367 #403

### DIFF
--- a/gwdetchar/lasso/__main__.py
+++ b/gwdetchar/lasso/__main__.py
@@ -34,7 +34,7 @@ from sklearn.preprocessing import scale
 from sklearn.preprocessing import *
 
 from pandas import set_option
-
+import pandas as pd
 from gwpy.detector import ChannelList
 from gwpy.io import nds2 as io_nds2
 
@@ -617,7 +617,9 @@ def main(args=None):
     nafter = len(auxdata)
     LOGGER.debug('Removed {0} channels with bad data'.format(nbefore - nafter))
     LOGGER.debug('{0} channels remaining'.format(nafter))
-    data = numpy.array([scale(ts.value) for ts in auxdata.values()]).T
+    data = numpy.array([ts.value for ts in auxdata.values()]).T
+    scaler = StandardScaler()
+    data = scaler.fit_transform(data)
 
     # -- perform lasso regression -------------------
 
--- a/gwdetchar/lasso/__main__.py
+++ b/gwdetchar/lasso/__main__.py
@@ -34,7 +34,6 @@ from sklearn.preprocessing import scale
 from sklearn.preprocessing import *
 
 from pandas import set_option
-import pandas as pd
 from gwpy.detector import ChannelList
 from gwpy.io import nds2 as io_nds2
 
--- a/gwdetchar/lasso/__main__.py
+++ b/gwdetchar/lasso/__main__.py
@@ -594,9 +594,9 @@ def main(args=None):
         frametype=frametype, nproc=args.nproc, pad=0).crop(start, end)
 
     # re-order aux channels to the same order as channel list
-    
+
     [auxdata.move_to_end(key=ch) for ch in channels if ch]
-        
+
     # -- removes flat data to be re-introdused later
 
     LOGGER.info('-- Pre-processing auxiliary channel data')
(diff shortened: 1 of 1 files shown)


lukecotnoir:
The newest commit is fixing some issues which caused errors to be thrown. 

1. The multithreading pool used in Lasso was never closed causing a 'bad file descriptor' error after Lasso was run. Closing the thread pool resolved the issue.
2. When converting the results to pandas, setting the max_colwidth option to -1 is deprecated so None should be used instead.
3. Setting various options for axes in matplotlib can no longer be done inline with plot.gca(), so they should be done seperately calling setters for the axis object.

lukecotnoir:
Committed a change fixing the sklearn scaling done to the aux channel data before fitting. Before, each channel was scaled using the general scale() function and each channel threw an error saying to "pre-scale your features". To fix this issue, I instead used the StandardScaler.fit_transform() function to scale the numpy array holding all the aux channel data at once.

This produces the same final results and does not throw the errors.

gwdetchar/lasso/__main__.py
```diff
@@ -590,10 +594,14 @@ def main(args=None):
...
 
+    # re-order aux channels to the same order as channel list
+    
+    [auxdata.move_to_end(key=ch) for ch in channels if ch]
+        
     # -- removes flat data to be re-introdused later
 
     LOGGER.info('-- Pre-processing auxiliary channel data')
-
+
```
eagoetz, line 604:
```suggestion

```

gwdetchar/lasso/__main__.py
```diff
@@ -31,9 +31,9 @@
 
 from sklearn import linear_model
 from sklearn.preprocessing import scale
+from sklearn.preprocessing import *
```
eagoetz, line 34: It's a bit unusual to import with `*`. Perhaps better to just import what you need. This may help solve some flake8 errors

gwdetchar/lasso/__main__.py
```diff
@@ -590,10 +593,13 @@ def main(args=None):
         channels, start, end, verbose='Reading:'.rjust(30),
         frametype=frametype, nproc=args.nproc, pad=0).crop(start, end)
 
+    # re-order aux channels to the same order as channel list
+    
+    [auxdata.move_to_end(key=ch) for ch in channels if ch]
+
```
eagoetz, line 599:
```suggestion
    # re-order aux channels to the same order as channel list

    [auxdata.move_to_end(key=ch) for ch in channels if ch]

```

eagoetz (approved):
LGTM

PR #423, opened 2023-04-06
Add ER15 and O4 epochs

duncanmmacleod:
This PR adds GPS segments for the (current best-guess) ER15 and O4 epochs.

This is required to have `gwdetchar.omegao.batch` automatically use O4 accounting tags.


eagoetz:
@duncanmmacleod Thanks, sorry we missed this. I spotted this issue and got it updated in https://github.com/gwdetchar/gwdetchar/pull/425. We can close this PR since that fixed the same thing

PR #425, opened 2023-04-24
Fix for changed valid condor tags file, update O3/O4 segments

eagoetz:
This PR updates the valid condor tags file, which has changed format. It is no longer a json file and is named differently. Each line of the file looks like
`* ligo.sim.o5.sgwb.targeted_ld.stamp found`

In addition the segment list for predefined intervals like O3 and O4 is updated


gwdetchar/const.py
```diff
@@ -46,7 +46,8 @@
     ('ER9', Segment(1152136817, 1152255617)),
     ('ER10', Segment(1161907217, 1164499217)),
     ('O2', Segment(1164499217, 1187733618)),
-    ('O3', Segment(1187733618, 2000000000)),
+    ('O3', Segment(1187733618, 1269363618)),
+    ('O4', Segment(1368982818, 2000000000)),
```
robertbruntz, line 50: Do we care that GPS time 1368982818 = Wed May 24 17:00:00 UTC 2023, not 15:00 UTC?  Also, should there be one for ER15?  (I suspect not, since there isn't one for ER11-14; just checking.)

gwdetchar/tests/test_condor.py
```diff
@@ -46,7 +45,7 @@ def test_accounting_epoch(gps, epoch):
 def tagfile():
     name = tempfile.mktemp()
     with open(name, 'w') as tmp:
-        json.dump({'groups': ['tag1', 'tag2', 'tag3']}, tmp)
+        tmp.write("* tag1 found\n* tag2 found\n* tag3 found\n")
```
robertbruntz, line 48: This change in formatting (not the change in how it's saved) versus the old formatting is intentional, right?

gwdetchar/const.py
eagoetz: Good catch, I'll fix the O4 one. This was a copy and paste error. I was unsure about the ER15 segment since there was none of the ER11-14, as you already mentioned. It couldn't hurt to add it. I will do it.

gwdetchar/tests/test_condor.py
eagoetz: Yes, the file format is completely different now. The list of valid tags is not a json file but rather an ASCII file with entries like: `* tag1 found`

robertbruntz: Never mind - I just noticed that it's in the PR's actual description.

areeda (approved):
let's try it.

gwdetchar/const.py
```diff
@@ -46,7 +46,9 @@
     ('ER9', Segment(1152136817, 1152255617)),
     ('ER10', Segment(1161907217, 1164499217)),
     ('O2', Segment(1164499217, 1187733618)),
-    ('O3', Segment(1187733618, 2000000000)),
+    ('O3', Segment(1187733618, 1269363618)),
+    ('ER15', Segment(1269356418, 1368975618)),
+    ('O4', Segment(1368975618, 2000000000)),
```
robertbruntz, line 51: So the standard we're going with is 'actual end of one run = figurative start of the next run'?  Thus, O3 starts at GPS time 1187733618 = Fri Aug 25 22:00:00 UTC 2017 (actual end of O2 [verified]) and ends at GPS time 1269363618 = Fri Mar 27 17:00:00 UTC 2020 (actual end of O3 [verified]).  I think ER15 should start at GPS time 1269363618 = Fri Mar 27 17:00:00 UTC 2020 then, not GPS time 1269356418 = Fri Mar 27 15:00:00 UTC 2020, which assigns 15:00-17:00 to both O3 and ER15.

eagoetz: I see, a bit strange but ok

eagoetz: How about O3 runs until ER15 starts. Meaning, 
```
    ('O2', Segment(1164499217, 1187733618)),
    ('O3', Segment(1187733618, 1269363618)),
    ('ER15', Segment(1269363618, 1368975618)),
    ('O4', Segment(1368975618, 2000000000))
```

eagoetz: The conventions for epochs in this list are actually quite mixed up.

robertbruntz: They are a bit mixed up, and it seems odd to me to have to extend O3 forward or ER15 backward to cover time they didn't actually include.  Do we know what actually cares about those epoch time spans?  If this is an urgent PR, I think your current setting, with ER15 starting at the *actual* end of O3 (similar to how the O3 epoch starts at the *actual* end of O2), is fine, and we can ask Duncan and/or open a ticket to decide what it *should* be and if anything needs to be changed.

eagoetz: I agree. Not all GPS times are covered by earlier epochs either. I'm ok with epochs starting when we actually start the epoch and allowing the end to run up until the start of the next epoch. We can figure it out later though and make a better choice going forward. Perhaps that was decided before Alex left. I am not sure

robertbruntz:
LGTM

robertbruntz (approved):
LGTM

PR #427, opened 2023-04-25
Fix HTML CI test

eagoetz:
This PR fixes the HTML unit tests because different `pygments` versions emit slightly different output. Hopefully this is more robust to different versions of `pygments`

These CI tests used pygments-2.13. My local version is 2.15 and also ran fine. Hopefully this solves the problem


areeda (approved):
LGTM

PR #428, opened 2023-04-26
Fix lasso unit test find_alpha

eagoetz:
The Lasso CI had an exact floating point comparison `==`. This is too strict and is relaxed using `nptest.assert_almost_equal()`


robertbruntz (approved):
LGTM.

You might want to leave a note or link to an issue as to why this was necessary, but that's up to you.

PR #430, opened 2023-04-28
Bugfix to lasso table display

andrew-lundgren:
Pandas 2.0.1 doesn't allow nonnegative integers for col_width. This was causing lasso to fail. I think `None` is probably the equivalent option. I tested this with a short run which produced an output page.

### DIFF
--- a/gwdetchar/lasso/__main__.py
+++ b/gwdetchar/lasso/__main__.py
@@ -640,7 +640,7 @@ def main(args=None):
     print('\n\n')
 
     # convert to pandas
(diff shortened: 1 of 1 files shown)


marissawalker:
This seems like a good fix to me!

marissawalker:
I think this same change was made in the latest pull request that was just merged, so we could probably close this pull request

eagoetz:
@andrew-lundgren @marissawalker closing this PR since it was fixed in #422

PR #431, opened 2023-05-16
Fix import of omega.batch

eagoetz:
Importing omega.batch on the cluster causes it to look for the valid tags. The old version defaulted to GPS time = 0, which would return the "S6" tag that is no longer valid on LDG computing clusters. This PR fixes the import so that it will be the GPS time that the module is imported as a default rather than GPS = 0

Also update some strings using format() to f-strings


areeda (approved):
LGTM

PR #432, opened 2023-05-17
Remove ER15 epoch

eagoetz:
The ER15 epoch will generate invalid condor accounting tags, so we can only use O3 or O4 until ERxx is allowed as a valid accounting tag


areeda (approved):
LGTM

PR #433, opened 2023-06-07
Skip processing of missing channels

dethodav:
This PR updates `gwdetchar-omega` so that when data from channels is not found, the channel is simply skipped rather than returning an error. 

An example omegascan with this new update is here: https://ldas-jobs.ligo.caltech.edu/~derek.davis/detchar/O4/gwdetchar_tests/omega_missing_chans/

The output (with a message showing some channels were skipped) is saved here: https://ldas-jobs.ligo.caltech.edu/~derek.davis/detchar/O4/gwdetchar_tests/omega_missing_chans/omega_output.txt

The above links are being populated when this PR is created, I'll add a comment once these are complete.

### DIFF
--- a/gwdetchar/omega/__main__.py
+++ b/gwdetchar/omega/__main__.py
@@ -386,8 +386,7 @@ def main(args=None):
                 continue
 
             if channel.name not in data:
-                LOGGER.warning(
-                    ' -- No data found for {}, skipping channel'.format(channel.name))
+                # Channel must not have been found by get_data()
                 continue
 
             analyzed = _scan_channel(
--- a/gwdetchar/omega/__main__.py
+++ b/gwdetchar/omega/__main__.py
@@ -385,6 +385,11 @@ def main(args=None):
                 html.write_qscan_page(ifo, gps, analyzed, **htmlv)
                 continue
 
+            if channel.name not in data:
+                LOGGER.warning(
+                    ' -- No data found for {}, skipping channel'.format(channel.name))
+                continue
+
             analyzed = _scan_channel(
                 channel, data[channel.name], analyzed, gps, block,
                 args.far_threshold, (args.frequency_scaling == 'log'),


eagoetz:
@dethodav Thanks for putting something together.

I'm curious though: it seems like the method `gwdetchar.io.datafind.get_data()` should perhaps not return a channel (but maybe it currently does? Can you check this?) that does not exist in frames. If it does not return the data for non-existent channels, then perhaps a more straightforward solution is to take the next loop `for channel in block.channels:` to be a loop over each channel in the `data` variable.

dethodav:
@eagoetz `gwdetchar.io.datafind.get_data()` correctly ignores any missing channels, but the list of missing channels isn't propagated into the next for loop. Changing the `for channel in block.channels:` loop would require less code, but then there wouldn't be any indication in the log messages that no data was found for a channel.

dethodav:
The example omegascan is now complete, and correctly skips missing channels. For example, in the log we can see
```
gwdetchar-omega 2023-06-07 13:05:00 PDT  WARNING:  -- No data found for L1:PEM-CS_MAG_EBAY_SUSRACK_QUAD_SUM_DQ, skipping channel
gwdetchar-omega 2023-06-07 13:05:00 PDT  WARNING:  -- No data found for L1:PEM-CS_MAG_LVEA_OUTPUTOPTICS_QUAD_SUM_DQ, skipping channel
gwdetchar-omega 2023-06-07 13:05:00 PDT  WARNING:  -- No data found for L1:PEM-CS_MAG_LVEA_VERTEX_QUAD_SUM_DQ, skipping channel
gwdetchar-omega 2023-06-07 13:05:00 PDT  WARNING:  -- No data found for L1:PEM-EX_MAG_EBAY_SUSRACK_QUAD_SUM_DQ, skipping channel
gwdetchar-omega 2023-06-07 13:05:00 PDT  WARNING:  -- No data found for L1:PEM-EX_MAG_VEA_FLOOR_QUAD_SUM_DQ, skipping channel
gwdetchar-omega 2023-06-07 13:05:00 PDT  WARNING:  -- No data found for L1:PEM-EY_MAG_VEA_FLOOR_QUAD_SUM_DQ, skipping channel
gwdetchar-omega 2023-06-07 13:05:00 PDT  WARNING:  -- No data found for L1:PEM-EY_VAULT_MAG_COIL_Z_DQ, skipping channel
```

eagoetz:
My suggestion would be for the message to be emitted in `get_data()` and then loop over the valid channels. Is this possible?

dethodav:
Ok, I think I understand what happens - there IS a warning displayed when a channel isn't found, but it's output to standard error instead of standard output. A copy of my standard error for this run is here: 

https://ldas-jobs.ligo.caltech.edu/~derek.davis/detchar/O4/gwdetchar_tests/omega_missing_chans/omega_err.txt

For example, the error messages contain 
```
/home/derek.davis/src/gwdetchar/gwdetchar/io/datafind.py:117: UserWarning: L1:PEM-CS_MAG_EBAY_SUSRACK_QUAD_SUM_DQ is being removed because it was not available in all requested files
```

Since an error message IS being displayed, I think that @eagoetz's solution makes the most sense - I'll adjust the PR accordingly.

eagoetz:
@dethodav Ping on this PR.

I am thinking to push out a release tomorrow and then subsequently test and update the summary environments after testing the changes needed for lasso updates.

Thanks!

dethodav:
@eagoetz I'm looking at this again, and I realize that we can't change the `for channel in block.channels:` to instead loop over the channels in the `data` variable because the keys of `data` are strings rather than `Channel` objects. 

I can still remove the duplicate error message (on L390) - would that be sufficient?

eagoetz:
@dethodav Thanks for the update!

eagoetz (approved):
LGTM

PR #438, opened 2023-06-20
Update condor workflow for batch omega scans

duncanmmacleod:
This PR updates the condor workflow generated and used in gwdetchar-omega-batch:

- use condor file transfer as much as possible, this should make the workflow more resilient
- specify `request_cpus`, `request_disk`, and `request_memory` explicitly
- replace `getenv = true` with specific minimum set of environment variables

This has been tested on the LDAS-LLO system via

```shell
python3 -m gwdetchar.omega.batch 1371255906.044922 -o test -s -j 1 -i L1 --config-file L-L1_R-selected.ini
```

and a few other variants.

### DIFF
--- a/gwdetchar/io/html.py
+++ b/gwdetchar/io/html.py
@@ -1264,18 +1264,23 @@ def package_list():
     """
     prefix = sys.prefix
     if (Path(prefix) / "conda-meta").is_dir():
-        raw = subprocess.check_output(
-            ["conda", "list",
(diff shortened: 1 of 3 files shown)


eagoetz:
@duncanmmacleod Thanks for putting this together. I wonder if the CI failures will be fixed with a rebase. Could you check where this stands? Thank you!

eagoetz (approved):
LGTM

PR #439, opened 2023-06-22
Move daily-lasso code into gwdetchar.lasso

eagoetz:
This PR moves scripted code for daily running of lasso into gwdetchar.lasso.__main__ so that there is less duplication of effort and code is maintained in a singular location. It also enables more people to use lasso over multiple segments, which is essentially what the daily-lasso code was doing

An example output is here: https://ldas-jobs.ligo-wa.caltech.edu/~evan.goetz/gwdetchar/lasso/

Compare this to the current production: https://ldas-jobs.ligo-wa.caltech.edu/~detchar/lasso/summary/day/20230622/

### DIFF
--- a/gwdetchar/lasso/__main__.py
+++ b/gwdetchar/lasso/__main__.py
@@ -16,12 +16,12 @@
 # You should have received a copy of the GNU General Public License
 # along with GW DetChar.  If not, see <http://www.gnu.org/licenses/>.
 
+import fnmatch
 import multiprocessing
 import os
 import re
 import sys
-
-from gwpy.timeseries import TimeSeries
+from getpass import getuser
 
 import numpy
 
@@ -33,8 +33,14 @@ from sklearn import linear_model
 from sklearn.preprocessing import StandardScaler, scale
 
 from pandas import set_option
-from gwpy.detector import ChannelList
-from gwpy.io import nds2 as io_nds2
+from gwpy.io.datafind import find_urls
+from gwpy.io.gwf import get_channel_names
+from gwpy.segments import (
+    SegmentList,
+    DataQualityFlag,
+)
+from gwpy.time import tconvert
+from gwpy.timeseries import TimeSeries
 
 from .. import (cli, lasso as gwlasso)
 from ..io.datafind import get_data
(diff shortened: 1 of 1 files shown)


marissawalker:
Thanks Evan! Overall this looks really good to me. It will greatly simplify daily-lasso and fixes the problem that we had where running lasso without a channel list didn't work. I was also able to run the new code with no issues. I just have a couple of minor concerns:

- Now the only flag options for running lasso would be during lock-nominal or analysis-ready. While those are going to be the most common use cases, I wonder if there should be an option to not have a flag (and just run on the exact segment the user puts in), or to pass a custom flag. For example, if someone wanted to run lasso with a primary channel that isn't range (e.g. some subsystem investigation), and they wanted to look at it during time when the detector isn't locked but that channel is still taking data. This is probably not going to cause immediate issues for anyone but I think it is important to fix this so that the code is flexible enough for us to do other investigations.
- With this version, if we run over a single segment, it will now create a summary html page and then you have to click "full results" to see the page like it was before. The extra click might not be a big deal but I did want to note that this would be a slight change to the user's experience doing a manual run. 

Neither of those concerns would make a difference on the daily lasso runs, and probably wouldn't make a difference to most people, so if they're not easy to change, I think it's probably fine to go ahead and merge, but note for later that we might want to tweak things (especially the segment option).

eagoetz:
@marissawalker Thanks for the feedback! Sounds like we could proceed with this PR without further changes. If I may provide some comments to your feedback:
> Now the only flag options for running lasso would be during lock-nominal or analysis-ready. While those are going to be the most common use cases, I wonder if there should be an option to not have a flag (and just run on the exact segment the user puts in), or to pass a custom flag. For example, if someone wanted to run lasso with a primary channel that isn't range (e.g. some subsystem investigation), and they wanted to look at it during time when the detector isn't locked but that channel is still taking data. This is probably not going to cause immediate issues for anyone but I think it is important to fix this so that the code is flexible enough for us to do other investigations.

I believe this would require the state flag to become an optional argument and remove the boolean logic of the low-noise state flag. If a state flag is not given, then lasso would just try to process the data in the segment interval requested. I suggest this be put into a separate pull request if you want to go in that direction.

> With this version, if we run over a single segment, it will now create a summary html page and then you have to click "full results" to see the page like it was before. The extra click might not be a big deal but I did want to note that this would be a slight change to the user's experience doing a manual run.

That's true, the original way that lasso runs in gwdetchar is over a single segment making the segment pages. The daily-lasso script was made so that the summary page for the day (or interval) could handle multiple segments. To me, the latter should be the "normal" operation whether running over single or multiple segments.

eagoetz:
@areeda I think this PR would possibly impact how LDVW works with Lasso. I'm not sure how you want to proceed, but I'd be happy to walk you through these changes and assist with any LDVW changes that would be required. Thanks

marissawalker:
I'm fine with this being merged now, just as long as people are aware someone might need to change the flag options later.

PR #442, opened 2023-08-09
Fix command line output for main lasso page

eagoetz:
This PR updates the command line text printed to the summary page so that it shows
```
$ python -m gwdetchar.lasso
```
instead of
```
$ __main__.py
```


areeda (approved):
lgtn

PR #443, opened 2023-08-21
Fix "top channels" blue ribbon dropdown

eagoetz:
This PR fixes an issue that the blue ribbon dropdown showing additional plots for the top channels was not working. It is fixed with this small typo fix


areeda (approved):
lgtm

PR #444, opened 2023-09-20
Allow generic file names when using FancyPlot

### LINKED ISSUE #415: gwdetchar.io.html.fancybox_img() requires filenames to be specifically formatted
Code in `gwdetchar.io.html.fancybox_img()` requires filenames to be formatted a specific way. For example, it expect the filename to be formatted with at least one `-`. The function then specifically assumes that the strings around the `-` are the channel name, and the duration is the last string element. The `channel` and `duration` values are used as `id_` values.

Right now the function is too specific and cannot be used with generic filenames. A possible solution is to pass optional channel name and duration as arguments to the function. If they are not given, then use some different values for the `id_`

dethodav:
This MR allows users to use the `io.html.fancybox_img()` function with images that do not follow the "standard" naming convention. Although most tools follow this convention, requiring a naming convention has led to challenges with using `gwdetchar` in more generic applications (such as the IGWN DQR).

Closes #415

### DIFF
--- a/gwdetchar/io/html.py
+++ b/gwdetchar/io/html.py
@@ -768,7 +768,9 @@ def fancybox_img(img, linkparams=dict(), lazy=False, **params):
     substrings = os.path.basename(img).split('-')
     try:
         # If is the expected format, use channel and duration
-        if substrings[0] not in OBSERVATORY_MAP.keys():
+        ifo_str = substrings[0]
+        if not (len(ifo_str) == 2 and ifo_str[0].isalpha() and 
+                ifo_str[1].isdigit()):
             raise TypeError
         channel = f'{substrings[0]}-{substrings[1]}'
         duration = int(substrings[-1].split('.')[0])
--- a/gwdetchar/io/html.py
+++ b/gwdetchar/io/html.py
@@ -769,13 +769,15 @@ def fancybox_img(img, linkparams=dict(), lazy=False, **params):
     try:
         # If is the expected format, use channel and duration
         ifo_str = substrings[0]
-        if not (len(ifo_str) == 2 and ifo_str[0].isalpha() and 
+        if not (len(ifo_str) == 2 and ifo_str[0].isalpha() and
                 ifo_str[1].isdigit()):
             raise TypeError
+        duration = substrings[-1].split('.')[0]
+        if not duration.isdigit():
+            raise TypeError
         channel = f'{substrings[0]}-{substrings[1]}'
-        duration = int(substrings[-1].split('.')[0])
         img_str = f'{channel}_{duration}'
-    except:
+    except TypeError:
         # If unexpected format, use entire image name
         # This only changes the label of the image in the html code
         # and not the format of the page itself
--- a/gwdetchar/io/html.py
+++ b/gwdetchar/io/html.py
@@ -766,9 +766,15 @@ def fancybox_img(img, linkparams=dict(), lazy=False, **params):
     aparams.update(linkparams)
     img = str(img)
     substrings = os.path.basename(img).split('-')
-    channel = '%s-%s' % tuple(substrings[:2])
-    duration = substrings[-1].split('.')[0]
-    page.a(href=img, id_='a_%s_%s' % (channel, duration), **aparams)
+    if len(substrings) == 4:
+        # this is the expected format, use channel and duration
+        channel = '%s-%s' % tuple(substrings[:2])
+        duration = substrings[-1].split('.')[0]
+        img_str = '%s_%s' % (channel, duration)
+    else:
+        # unexpected format, use entire image name
+        img_str = os.path.basename(img).split('.')[0]
+    page.a(href=img, id_='a_%s' % img_str, **aparams)
     src_attr = lazy and 'data-src' or 'src'
     imgparams = {
         'alt': os.path.basename(img),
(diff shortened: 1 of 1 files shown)


dethodav:
The motivation for this change is an issue in the (LVK-internal) DQR git repository: https://git.ligo.org/detchar/dqrtasks/-/issues/51

I have not yet heard whether this fix fully resolves the relevant problem from the reporting user.

eagoetz:
@dethodav Thanks for the PR, I only just now saw this on my radar. Is it ready for review?

If yes, I have two minor suggestions:
1. While we are at it modifying these lines, would you be willing to change them to use f-strings instead of the old style using `%`
2. I am a little concerned that someone may accidentally use this function with a plot having a name with four fields separated by `-` but are not the same "standard" format. Would you be willing to add a few checks to guarantee the fields look like `<IFO>-<channel/plot descriptor>-<start-time>-<duration>`? Maybe just confirm that the first field looks like an IFO name, the third field is an int and the last is also an int?

Thoughts?

dethodav:
@eagoetz 

1. Yes, I'd be happy to make that change
2. What is your concern with someone using 4 fields that aren't the "right" values? I can see why it would make sense to check that, but in the current code, it's allowed for someone to use 4 fields that don't follow this format without issues.

eagoetz:
>     2. What is your concern with someone using 4 fields that aren't the "right" values? I can see why it would make sense to check that, but in the current code, it's allowed for someone to use 4 fields that don't follow this format without issues.

I concur that the code works for the 4 fields that don't follow the same format of those 4 fields, but maybe if the code is trying to assume something about the content of those fields (like the start time and duration), maybe just fall back to the alternative. To me the code would behave more sensibly in that case. I don't have strong opinions about it, so if you'd rather leave it as-is, we can just move along and worry about it some other time.

dethodav:
@eagoetz That makes sense - I wasn't sure if there was an issue with the code I wasn't seeing. I'll go ahead and add that check along with updating the formatting.

eagoetz:
@dethodav In addition to the suggested updates to the code, you should also rebase as I've fixed the documentation CI failure in a different PR that was merged

dethodav:
@eagoetz This should now be updated

eagoetz:
@dethodav You may have already figured out why the CI build is failing, but in case you haven't had a chance to take a look: it looks like the test for whether the first field is an observatory map field means that `X1` will fail the test, so it falls back to the file string, which doesn't match the test condition. 

For example, the expected string pattern is `img id="img_X1-TEST_AUX_4"`, but what this version of the code does is generate `img id="img_X1-TEST_AUX-test-4"`.

One suggestion would be to relax the conditions somewhat: test the first field is two characters (alphabetical + numeric) and the last field is numeric only. (Note: I would have normally suggested to test the last two fields for integers, but this wouldn't pass the CI build either.)

gwdetchar/io/html.py
```diff
@@ -766,17 +766,31 @@ def fancybox_img(img, linkparams=dict(), lazy=False, **params):
...
     img = str(img)
     substrings = os.path.basename(img).split('-')
-    channel = '%s-%s' % tuple(substrings[:2])
-    duration = substrings[-1].split('.')[0]
-    page.a(href=img, id_='a_%s_%s' % (channel, duration), **aparams)
+    try:
+        # If is the expected format, use channel and duration
+        ifo_str = substrings[0]
+        if not (len(ifo_str) == 2 and ifo_str[0].isalpha() and
+                ifo_str[-1].isdigit()):
```
eagoetz, line 773: Just a quick question: why do you use `ifo_str[-1]` instead of `ifo_str[1]`? It looked like the comment string from your last commit was for something about `len(ifo_str) = 1` case, but doesn't the `len(ifo_str) != 2` case already trigger the `TypeError` exception?

dethodav: This issue is that all of the conditions are checked before evaluating the `if` statement, rather than checking the conditions sequentially. In the case of a 1-length string, `len(ifo_str) == 2` would evaluate to `False`, but  `ifo_str[1].isdigit()` would return an `IndexError`. I could add a catch for this, but decided to just test  `ifo_str[-1].isdigit()`, which would not encounter an error for a 1-length string.

eagoetz: Are you sure that python doesn't see the first argument as `False` in the if statement and then automatically moves on to the else? I am almost sure that I've relied on length checking as the first in a list of conditionals and it works. I'll do a quick check with some code...

eagoetz: For example:
```python
>>> ifo_str = 'H'
>>> if not (len(ifo_str) == 2 and ifo_str[0].isalpha() and ifo_str[1].isdigit()):
...   print('foo')
... else:
...   print('bar')
... 
foo
>>> if not (len(ifo_str) == 2 and ifo_str[0].isalpha() and ifo_str[1].isdigit()):
...   print('foo')
... else:
...   print('bar')
... 
bar
```
Seems to work for me

dethodav: Thanks for the example @eagoetz, I misunderstood how these compound `if` statements were being tested. I just pushed your suggestion.

eagoetz: @dethodav Thanks! It wasn't a big deal, but I had just noticed your changes and wondered if I was missing something. Thanks for checking and making an adjustment that was not strictly required for the review. All good though, thanks for everything!

eagoetz (approved):
LGTM

PR #445, opened 2023-11-16
Exclude OAF-CAL_RANGE in addition to OAF-RANGE channels from LASSO

eagoetz:
This PR is a minor update to the channel exclusion list so that the recently added OAF-CAL_RANGE channels at L1 are excluded


areeda (approved):
LGTM

PR #447, opened 2023-12-15
LASSO: intersect data segments with DQ flag segments

eagoetz:
This PR adds the capability of LASSO to intersect the DQ segment list with the actually available data. This makes the program more robust to gaps in frame availability


areeda (approved):
:ppks Good

PR #448, opened 2023-12-18
Fix omega.batch condor submission problems

eagoetz:
This PR fixes the problem that condor_submit_dag filters the environment variables unless they are specified to be passed, so even though the submit file for the jobs have been given `getenv = ...` they will not return anything other than empty unless the dag also has passed the variables.


areeda (approved):
LGTM

PR #449, opened 2023-12-22
Fix lasso channel exclusions

eagoetz:
This PR resolves LASSO channel exclusion pattern matching


eagoetz:
I can confirm these small modifications fix the channels that were supposed to be excluded actually do get excluded

areeda (approved):
LGTM

PR #450, opened 2024-01-27
Use on_gaps='warn' for lasso calls to get_data()

eagoetz:
This PR applies the use of `on_gaps='warn'` to all calls to `get_data()`. This avoids problems if some of the data is inaccessible even though the segment list says data is supposed to exist.

### DIFF
--- a/gwdetchar/lasso/__main__.py
+++ b/gwdetchar/lasso/__main__.py
@@ -296,7 +296,8 @@ def get_primary_ts(channel, start, end, filepath=None,
         LOGGER.info('Querying primary channel')
         return get_data(channel, start, end,
                         verbose='Reading primary:'.rjust(30),
-                        frametype=frametype, source=cache, nproc=nproc)
+                        frametype=frametype, source=cache, nproc=nproc,
+                        on_gaps='warn')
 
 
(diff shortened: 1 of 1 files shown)


gwdetchar/lasso/__main__.py
```diff
@@ -575,47 +577,56 @@ def main(args=None):
...
+        data_segs &= segs  # intersect with state segments
+        data_segs.coalesce()  # just in case
+    else:
+        data_segs = segs
+
+    # So far we're just warning if there are missing data segments, but we
+    # don't want to silently proceed if there is a significant gap. We raise
+    # an error if the gap is greater than 10 minutes = 600 seconds
+    if abs(segs) - abs(data_segs) > 600:
+        raise RuntimeError("Missing more than 600 s of frame data")
```
eagoetz, line 614:
```suggestion
    if abs(abs(segs) - abs(data_segs)) > 600:
        raise RuntimeError(f"Difference in state segs = {abs(segs)} s "
                           f"and data segs = {abs(data_segs)} s is > 600 s")
```

areeda (approved):
LGTM

PR #453, opened 2024-04-17
Fix dropdown alignment for menus that are right aligned

eagoetz:
This PR fixes an issue that Bootstrap 5 changed the name of `dropdown-menu-right` to `dropdown-menu-end`. I have searched and replaced those instances.


areeda (approved):
Looks good

PR #457, opened 2024-07-19
Update versions of github actions except for codecov requiring token

eagoetz:
This PR updates the version pins for the github actions except for codecov because there is a requirement to get a token. See https://github.com/codecov/codecov-action/blob/main/README.md#usage


eagoetz:
Ping @duncanmmacleod for review. Thanks!

PR #459, opened 2024-07-19
Pin to numpy versions less than v2

eagoetz:
Temporary change until other dependencies are updated to handle the new numpy version 2


areeda (approved):
LGTM

PR #461, opened 2024-11-04
Fix bug with dropdown menu when using bootstrap 5

dethodav:
This PR fixes a bug that was introduced when switching to bootstrap 5. At present, dropdown menus do not work on mobile platforms or for small windows. This was due to a parameter name change in bootstrap 4>5 not being applied to `gwdetchar`. 

Here is an example of the correct parameter name (`data-bs-target`) in the [bootstrap docs](https://getbootstrap.com/docs/5.0/components/dropdowns/#dark-dropdowns).

This is only a one-line change (and one line changed in the test HTML page).


eagoetz:
@dethodav Thanks, I thought I had fixed all those but apparently not. This one slipped through. Please rebase since I merged #457

eagoetz (approved):
LGTM

PR #462, opened 2024-12-05
Remove default segment server url and use gwpy default host

eagoetz:
This PR removes the default segment server and old S6 segment server (doesn't even exist anymore) and instead uses the defaults from gwpy


eagoetz:
Ping @duncanmmacleod can you review this PR? Thanks in advance!

duncanmmacleod (approved):
LGTM

PR #464, opened 2025-01-06
Remove unused ligo-segments requirement

duncanmmacleod:
This PR removes `ligo-segments` from the `dependencies` for this project, this package is never imported.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,6 @@ dependencies = [
   "gwpy >=3.0.0",
   "gwtrigfind",
   "lalsuite",
-  "ligo-segments",
   "lxml",
   "MarkupPy >=1.14",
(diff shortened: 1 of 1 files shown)


eagoetz:
@duncanmmacleod Seems reasonable. Probably this was left over from when there may have been some use for it in the past. We may need to make sure that there's no downstream packages that are expecting `ligo-segments` to be installed with `gwdetchar`

duncanmmacleod:
> @duncanmmacleod Seems reasonable. Probably this was left over from when there may have been some use for it in the past. We may need to make sure that there's no downstream packages that are expecting `ligo-segments` to be installed with `gwdetchar`

It shouldn't be the job of this project to provide transitive dependencies for other projects. If this is removed the others can (and should) adapt pretty quickly.

eagoetz (approved):
LGTM

PR #463, opened 2024-12-12
Update build.yml github actions

eagoetz:
This PR removes mamba with conda-incubator/setup-miniconda@v3. This was causing broken CI workflows

Issue #418, opened 2022-11-14
Download package list as yaml


dethodav:
An extremely helpful feature would be a button on the "about page" that allows the user to download the package list as a yaml file instead of a csv file. This would allow the user to easily recreate the package list in conda.

PR #465, opened 2025-01-06
Update setup-miniconda configuration

duncanmmacleod:
This PR updates the configuration for the `setup-miniconda` CI step.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,11 +60,9 @@ jobs:
     - name: Configure conda
       uses: conda-incubator/setup-miniconda@v3
       with:
-        auto-update-conda: true
-        channels: conda-forge
+        activate-environment: test
+        miniforge-version: latest
         python-version: ${{ matrix.python-version }}
-        # this is needed for caching to work properly:
-        use-only-tar-bz2: true
 
     - name: Conda info
       run: conda info --all


eagoetz:
@duncanmmacleod I want to understand how this jives with #463. It seems to partially undo the changes I made. Thanks

duncanmmacleod:
@eagoetz, this change just ensures that miniforge is used, and not miniconda, or anything else.

eagoetz:
@duncanmmacleod Ok, understood, but why is this needed? It was working fine by using the "simpler" version I had in place? For example, you don't need to activate the `test` environment (https://github.com/conda-incubator/setup-miniconda?tab=readme-ov-file#environment-activation) and doesn't the conda-forge channel give you what you want anyway? Thanks in advance for clarifying

duncanmmacleod:
@eagoetz, all I know is that the current builds are failing without this, and working with it. For some reason the current miniconda (from `defaults`) distribution is pulling in a really old version of `pip2conda`, that doesn't work any more, but when you switch to `miniforge`, the latest version is installed. This can probably be resolved with a version pin, but that irks me...

duncanmmacleod:
It looks like the tests are now tripping over https://github.com/pygments/pygments/commit/0853028bfe180b7c56ef20502f44a5d8d340216c.

eagoetz:
@duncanmmacleod Ok, right, I removed caching and the `use-only-tar-bz2: true` in https://github.com/gwpy/gwsumm/blob/master/.github/workflows/build.yml but I didn't do it in the `gwdetchar` package yet.

eagoetz:
@duncanmmacleod I don't quite understand why the version of `pygments` in the CI test has changed the color scheme from `#bbbbbb` to `#BBB`. I looked at their changelog summary but I didn't find something listed there (but maybe I missed it).

Also, what do you think about just removing the cache and `use-only-tar-bz2: true` line? Similar to what was done in https://github.com/gwpy/gwsumm/pull/418

duncanmmacleod:
> @duncanmmacleod I don't quite understand why the version of `pygments` in the CI test has changed the color scheme from `#bbbbbb` to `#BBB`. I looked at their changelog summary but I didn't find something listed there (but maybe I missed it).

The pygments commit I listed includes a backwards-incompatible change to the HTML content written by that library.

> Also, what do you think about just removing the cache and `use-only-tar-bz2: true` line? Similar to what was done in [gwpy/gwsumm#418](https://github.com/gwpy/gwsumm/pull/418)

I don't understand why the caching should be removed, but removing the `use-only-tar-bz2: true` option is a requirement.

eagoetz:
@duncanmmacleod Ah, ok, sorry I missed that point! Literally just released...!

Regarding the caching: if you read the documentation comment https://github.com/conda-incubator/setup-miniconda?tab=readme-ov-file#caching-packages:
```
use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
```
So if you remove that line, then caching doesn't work anyway. So my thinking was if we remove this critical line, then why keep the whole caching component.

duncanmmacleod:
I moved the pygments report to an issue: https://github.com/gwdetchar/gwdetchar/issues/467

duncanmmacleod:
> So if you remove that line, then caching doesn't work anyway. So my thinking was if we remove this critical line, then why keep the whole caching component.

I am optimistic that the caching with `.conda` package files will be fixed at some point, so leaving the cache configuration in place means that _when_ that is fixed this project will benefit automatically.

duncanmmacleod:
@eagoetz, please see https://github.com/gwpy/gwsumm/pull/420#issuecomment-2578334136.

duncanmmacleod:
This was superseded by #468.

PR #466, opened 2025-01-06
Add dependabot configuration

duncanmmacleod:
This PR adds a [Dependabot configuration file](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates) to automatically update github actions versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+# Please see the documentation for all configuration options:
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/" # Location of package manifests
+    schedule:
+      interval: "weekly"
+    commit-message:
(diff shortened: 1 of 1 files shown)


eagoetz:
@duncanmmacleod How frequent do the github actions get new revisions such that this needs a weekly workflow?

duncanmmacleod:
@eagoetz, most of the actions get updated fairly rarely, so once a week is more than enough to catch updates. Given that only the major versions are referenced in the workflow configurations, and most of those are on v4/v5, that is roughly one major revision even year or two on average.

If you mean that weekly is too often to run, you shouldn't even notice it, since most of the time it won't result in anything.

eagoetz:
@duncanmmacleod Ok, sounds reasonable. What about the failed CI tests? Should we wait until that is resolved?

duncanmmacleod:
> @duncanmmacleod Ok, sounds reasonable. What about the failed CI tests? Should we wait until that is resolved?

See https://github.com/gwdetchar/gwdetchar/issues/467.

eagoetz (approved):
LGTM

PR #470, opened 2025-02-17
Revert "Pin to numpy versions less than v2 (#459)"

duncanmmacleod:
This reverts commit d76295d3d745386c624714f6d2f5749351bb6f8e.


eagoetz:
@duncanmmacleod sorry for the perhaps miscommunication, this is already addressed in https://github.com/gwdetchar/gwdetchar/pull/458 which is ready for your review

PR #480, opened 2025-05-28
Functions for DQ segs intersecting with data available segs

eagoetz:
This PR does several things:
- Moves `utils.py` to a `utils` submodule
- Adds `segments.py` to `utils` submodule with a new function `intersection_data_segs()` to perform the common task of taking the intersection of available frame files with data quality segments
- Moves utilities testing into `utils/tests/` and adds `utils/tests/test_segments.py`
- Change to use gwdatafind.find_urls() instead of through gwpy


areeda:
Looks good and makes sense

PR #485, opened 2025-09-24
Deal wuth Omega accounting-group-user pointing to detchar*

areeda:
While moving hveto from the local universe to the vanilla universe in certain situation it was complicated to pass the user as an environment variable. So this update detects when the calculated default user is detchar or detchar.dev


areeda:
Branched from updated fork resulted in 35 commits when only 34 were needed.

PR #494, opened 2025-12-04
Fix lasso auxiliary frametype variable name

eagoetz:
This PR fixes a problem where the `aux_frametype` is not part of the ArgumentParser namespace object. The problem was introduced in #480


areeda (approved):
LGTM

PR #499, opened 2026-01-22
Use match= with pytest.warns to assert warnings

duncanmmacleod:
This PR updates warning checks in tests to use the `match` keyword argument of `pytest.warns`.

The tests want to assert that a specific warning (identified by message) _was_ emitted, and not that _no other_ warnings were emitted - using `match` is the most elegant way to do that, and makes the code more robust to other warnings.

Also use it with `pytest.raises`.


eagoetz (approved):
LGTM

PR #500, opened 2026-02-27
Update CI/CD workflow to use uv instead of conda

duncanmmacleod:
This PR updates the CI/CD build workflow to use `uv` to install dependencies instead of `conda`.

`uv` is much faster, and this project doesn't actually need anything fancy.


eagoetz (approved):
Thanks @duncanmmacleod for the nice example of how to do this.

LGTM

PR #501, opened 2026-02-27
Explicitly control datetime format for HTML

duncanmmacleod:
This PR fixes an HTML formatting issue where a timezone-aware `datetime` has a different `str` representation to a naive `datetime`.

GWpy 4.0.0 returns aware `datetime`s, including timezone information; the easiest solution is to control it explicitly so there are no surprises.


eagoetz:
Superseded by #515

